PR #2, opened 2024-02-28
Use input_id to track inputs being accessed

cachafla:

- Fixes a bug where input metadata was not tracked correctly due to not relying on `input_id` for looking up inputs


even-steven (approved):

johnwalz97 (approved):
LGTM. Thanks for the fix 🙏

PR #8, opened 2024-03-06
Remove use of np.object

cachafla:

We have updated to a newer numpy and the error was thrown on our JupyterHub deployment. The error is self-explanatory:

```
`np.object` was a deprecated alias for the builtin `object`.
To avoid this error in existing code, use `object` by itself.
Doing this will not modify any behavior and is safe.
```


gtagle (approved):
LGTM

PR #12, opened 2024-03-09
Ensure all notebooks run end to end with new input/predictions syntax

cachafla:

- Ensure customer churn model runs end to end. Since this notebook has a complex use of inputs (3 datasets and 1 model) we now pass a per-test config to `run_documentation_tests()`
- Fix `FoundationModel` support
- Ensure all notebooks use `assign_predictions()` properly


validmind/models/foundation.py
```diff
@@ -59,9 +58,7 @@ def predict(self, X: pd.DataFrame):
         """
         Predict method for the model. This is a wrapper around the model's
         """
-        return np.array(
-            [self.predict_fn(self._build_prompt(x[1])) for x in X.iterrows()]
-        )
+        return [self.predict_fn(self._build_prompt(x[1])) for x in X.iterrows()]
```
cachafla, line 61: We didn't need to convert into numpy array since the `assign_predictions()` is already doing it.

validmind/vm_models/dataset.py
```diff
@@ -241,18 +244,19 @@ def assign_predictions(
...
-                self.__assign_prediction_values(model, pred_column, prediction_values)
+                return
+
+            logger.info("Running predict()... This may take a while")
+
+            # If the model is a FoundationModel, we need to pass the DataFrame to
+            # the predict method since it requires column names in order to format
+            # the input prompt with its template variables
+            x_only = (
+                self.x_df() if model.model_library() == "FoundationModel" else self.x
```
cachafla, line 255: @AnilSorathiya this was the issue with the foundation model tests: since they require column names to format prompts we're now passing a dataframe when the model is `FoundationModel`. Also, note that I simplified this code by using `self.x` which is already available.

validmind/vm_models/dataset.py
```diff
@@ -407,7 +411,7 @@ def y_pred(self, model_id) -> np.ndarray:
                 for name in self.columns
                 if name == self.prediction_column(model_id=model_id)
             ],
-        ]
+        ].flatten()
```
cachafla, line 414: This was key. We need to ensure we don't end up with numpy arrays inside numpy arrays.

cachafla:
This PR is ready for review. We still need a separate one to incorporate the doc updates made by @nrichers after we consolidated models-projects on the UI: https://github.com/validmind/developer-framework/pull/10

validmind/models/foundation.py
AnilSorathiya: 👍

validmind/vm_models/dataset.py
AnilSorathiya: thanks for this.

AnilSorathiya (approved):
Thanks for sorting this out. 👍

PR #13, opened 2024-03-11
[SC 3406] Follow up fix notebooks and tests after merging

AnilSorathiya:
## Internal Notes for Reviewer

Update text in the notebooks with some explanation using the `assign_predictions()` interface.


cachafla (approved):
👌

PR #15, opened 2024-03-13
Update release_notes_check.yamlRun release notes check only if not in draft mode

nrichers:

This PR makes a small tweak to our release notes check so that it runs only if a PR is not in draft mode. This should avoid frequent CI check failures when a PR is in progress.

Screenshot of show how this works in https://github.com/validmind/documentation/pull/159


cachafla (approved):
👌

PR #17, opened 2024-03-15
[SC 3568] Support for Polars dataframe

AnilSorathiya:

We are adding support to Polars Dataset/Dataframe as we support Pandas Dataset/Dataframe. 
ref: https://pola.rs/
- A customer should be able to use DataFrame from `Polars` library to create ValidMind dataset using the `init_dataset()` interface.
- All the tests should run seamlessly without any issues.


cachafla:
Skipping external notes until we have formal support for this dataset type.

cachafla (approved):
👌

PR #18, opened 2024-03-15
John6797/sc 3412/ability to compose a metric by combining

johnwalz97:

Initial support for composing Unit Metrics into Composite Metrics. This PR adds a new module `validmind.unit_metrics.composite.py` which adds support for running composite metrics (complex "test" made up of multiple individual unit metrics. These composite metrics can be created and run on the fly using the `run_test` function, saved to the backend as a metric_result object, and recreated dynamically from the template when running `run_documentation_tests()`. This works since when a result is logged, the individual unit metric IDs that make it up, along with the output template (if any), are saved as metadata attached to the content_id. The `content_id` starts with `validmind.composite_metric` so it can be identified as a composite metric at which time the associated metadata lets the dev framework rebuild and run it as its own "test".

I added on to @juanmleng 's notebook for running unit metrics to fully document this process end-to-end.

Check it out here:

https://github.com/validmind/developer-framework/blob/8430108c2adfdb5d0d5887edc314ca99c0292203/notebooks/how_to/run_unit_metrics.ipynb

## External Release Notes

ValidMind now supports the ability to "compose" multiple Unit Metrics into complex outputs. These composite metrics can be logged as a single result that can be used as a content block in your documentation and this result will be updated automatically when re-running the documentation test suite.


johnwalz97:
There's a few limitations with this initial pass at supporting composite metrics:

1. There's no text description for the metric result. We should look into creating one by default that combines the descriptions in some form from the unit metrics that make it up. This also means that things like `required_inputs` don't get bubbled up so a similar thing would need to happen for that to work.
2. There's no way to specify inputs or parameters for individual unit metrics. All parameters and inputs are defined at the composite metric level and get passed to each of the unit metrics.
3. There's no way to update the composite metric definition after it's created. The `update_metadata` function that's used to save the composite metric definition only sends data to the backend if the metadata doesn't exist. So you'd have to create a new composite metric with a new name. This should be an easy fix.
4. Finally, the unit metric results are not saved anywhere in the backend. I believe this will be done as part of a future ticket. For now unit metric results can only be saved as composite "normal" metric results.

cachafla:
Is it possible to get an automated summary when running a composite metric? Somehow "concatenating" all the individual summaries: https://github.com/validmind/developer-framework/blob/main/validmind/vm_models/test/unit_metric.py#L49-L61

I feel like requiring the user to always write an output template is gonna be a big challenge. Providing an output template should still be suggested as a way to customize and enhance the output though.

validmind/unit_metrics/composite.py
```diff
@@ -0,0 +1,159 @@
...
+
+class MetricProtocol(Protocol):
+    unit_metrics: list[str]
+
+    def output(self, results):
+        pass
+
+
+@dataclass
+class CompositeMetric(Metric):
```
cachafla, line 25: I suggest adding a simple description() implementation by default so that the user can take over on the UI and customize it.

cachafla:
> 1. There's no text description for the metric result. We should look into creating one by default that combines the descriptions in some form from the unit metrics that make it up. This also means that things like `required_inputs` don't get bubbled up so a similar thing would need to happen for that to work.

Right now the unit metrics don't have a description at all. I suggest adding a simple default description() for now so we have something to show.

> There's no way to specify inputs or parameters for individual unit metrics. All parameters and inputs are defined at the composite metric level and get passed to each of the unit metrics.

Suggestion: we could use the same approach as the `config` param for `run_documentation_tests`, but acknowledging that these params are meant for unit metrics. In this case it could work like this:

```py
result = run_test(
    name="ModelPerformance",
    unit_metrics=[
        "validmind.unit_metrics.sklearn.classification.F1",
        "validmind.unit_metrics.sklearn.classification.Precision",
        "validmind.unit_metrics.sklearn.classification.Recall",
        "validmind.unit_metrics.sklearn.classification.Accuracy",
        "validmind.unit_metrics.sklearn.classification.ROC_AUC"
    ],
    inputs=inputs,
    unit_metric_params={
      "validmind.unit_metrics.sklearn.classification.F1": {"average": "micro"}
    }
)
```

> There's no way to update the composite metric definition after it's created. The update_metadata function that's used to save the composite metric definition only sends data to the backend if the metadata doesn't exist. So you'd have to create a new composite metric with a new name. This should be an easy fix.

Can this be fixed in this PR?

> Finally, the unit metric results are not saved anywhere in the backend. I believe this will be done as part of a future ticket. For now unit metric results can only be saved as composite "normal" metric results.

Yep, future ticket is fine.

validmind/unit_metrics/composite.py
```diff
@@ -0,0 +1,159 @@
...
+    return class_def
+
+
+def run_metrics(
+    name: str = None,
+    metric_ids: list[str] = None,
+    output_template=None,
+    inputs=None,
+    params=None,
+    key: str = None,
```
cachafla, line 102: What's the need for a `key`? To allow a custom `test_id`? If that's the case I'd call it `test_id` so it's consistent with `run_test`.

johnwalz97:
> Is it possible to get an automated summary when running a composite metric? Somehow "concatenating" all the individual summaries: https://github.com/validmind/developer-framework/blob/main/validmind/vm_models/test/unit_metric.py#L49-L61
> 
> I feel like requiring the user to always write an output template is gonna be a big challenge. Providing an output template should still be suggested as a way to customize and enhance the output though.

Yeah this is the current behaviour. It automatically concatenates the results into a single-row table where the headers are the names of the unit metrics and the results are in the first row. The output template is only required if the user wants to customize it further

validmind/unit_metrics/composite.py
```diff
@@ -0,0 +1,243 @@
...
+            "__doc__": "Composite Metric built from multiple unit metrics",
+            "_unit_metrics": unit_metrics,
+            "_output_template": output_template,
+        },
+    )
+
+    required_inputs = set()
+    for metric_id in unit_metrics:
+        metric_cls = _get_metric_class(metric_id)
+        required_inputs.update(_extract_required_inputs(metric_cls))
```
johnwalz97, line 141: Let me know what you think of this solution @cachafla... The alternative would be to expect that unit metrics set the `required_inputs` property as part of the class definition.

johnwalz97:
Here is what it looks like with the default description:
image

johnwalz97:
> Is it possible to get an automated summary when running a composite metric? Somehow "concatenating" all the individual summaries: https://github.com/validmind/developer-framework/blob/main/validmind/vm_models/test/unit_metric.py#L49-L61
> 
> I feel like requiring the user to always write an output template is gonna be a big challenge. Providing an output template should still be suggested as a way to customize and enhance the output though.

This is what the automatic summary looks like (its just the unit metrics concatenated into a single table):
image

validmind/unit_metrics/composite.py
cachafla: Ah... I didn't notice the first set of unit metrics don't have `required_inputs` set. I think it's fine to expect the unit metrics to define this. We can fix the existing classes to have that information.

cachafla:
> > Is it possible to get an automated summary when running a composite metric? Somehow "concatenating" all the individual summaries: https://github.com/validmind/developer-framework/blob/main/validmind/vm_models/test/unit_metric.py#L49-L61
> > I feel like requiring the user to always write an output template is gonna be a big challenge. Providing an output template should still be suggested as a way to customize and enhance the output though.
> 
> This is what the automatic summary looks like (its just the unit metrics concatenated into a single table): image

Ah ok, awesome! Is it possible to concatenate "across" rows? i.e.

|Metric|Value|
|-|-|
|F1|0.8|
|AUC|0.7|

Once we add thresholds to unit metrics it's going to be easy to add them as columns. (obviously users can still manipulate the table with an `output_template`).

validmind/unit_metrics/composite.py
johnwalz97: Sounds good

cachafla (approved):
As a final request 🙂 can we remove the `output_template` from the notebook? The default one should be good enough for a demo.

PR #28, opened 2024-03-29
John6797/fix python dependencies

johnwalz97:

Python 3.11 is now supported! 🥳 

I opened this ticket to fix an issue where if you install the validmind package via pip tests will fail to load due to a dependency that made breaking changes via a minor version update. https://validmind.slack.com/archives/C04FZR7DHMK/p1711738162712649. One of our dependenices `ydata-profiling` uses the package `visions` which in turn uses a package `numba` for some of its optimizations. Numba made a breaking change in a minor version update and because that is accepted in `visions` requirements file, `pip` fails to realize that its not compatible. The simple fix for now is to just explicity tell `pip` to only install the working version...

But as part of this investigation, I looked into some of our dependencies and realized there were a few that could be updated or removed. And in particular, we could switch to `ydata-profiling` instead of `pandas-profiling` which lets us finally upgrade to Python 3.11. Next step is support 3.12 :)

To test these changes, I ran all of our unit tests and our integration notebooks on both python3.10 and python3.11.

## External Release Notes

Support for Python 3.11 has been added - users can now use the ValidMind Developer Framework with Python 3.8-3.11


pyproject.toml
```diff
@@ -15,40 +15,28 @@ version = "2.0.2"
 [tool.poetry.dependencies]
 aiohttp = {extras = ["speedups"], version = "^3.8.4"}
 arch = "^5.4.0"
-bert-score = "^0.3.13"
 catboost = "^1.2"
 click = "^8.0.4"
-datasets = {version = "^2.14.5", optional = true}
```
cachafla, line 21: @johnwalz97 why is this library being removed?

cachafla (approved):
Awesome 😎

PR #29, opened 2024-04-01
Custom Metric Function Decorator

johnwalz97:

This PR adds a new decorator thats simply called `metric`. This decorator can be used to register any function as a custom metric. This works because the decorator can dynamically create a Metric wrapper for any function by analyzing the function's signature to determine the inputs and parameters required to run the function. The `run` method for this Metric "wrapper" class calls the function with the correct inputs and parameters and then intelligently handles the values returned by the function. Any plotly or matplotlib objects are converted to ValidMind `Figure`s while lists or pandas dataframes are treated as summary tables. This works really well IMO and it means the only thing the user needs to worry about is using the VM model and dataset objects correctly and making sure that they only return supported objects.

I rewrote the custom tests notebook as part of this PR and tried to really clearly demonstrate how to build and use custom tests. I also moved the notebook into the code_samples directory so it will be accessible in the docs site.

I also added a new convenience function `reload` that will refresh the template to pull in any changes from the backend. This is helpful for notebook where the user is adding stuff to the template via the UI and then running documentation tests.

Check out the notebook here: https://github.com/validmind/developer-framework/blob/john6797/sc-3718/create-decorator-for-registering-one-off/notebooks/code_samples/custom_tests/implementing_custom_tests.ipynb

## External Release Notes

We've introduced a new `metric` decorator that's designed to turn any function into a ValidMind Metric. Custom metrics offer added flexibility by allowing users to extend the library of default metrics provided by ValidMind, enabling you to document any type of model or use case. This new decorator makes the process of creating and using custom metrics much simpler by almost completely eliminating the boilerplate code required to define and register a custom metric. Check out the [new notebook](https://docs.validmind.ai/notebooks/code_samples/custom_tests/implement_custom_tests.html) to see how you can start using custom metrics in your documentation!


cachafla (approved):
Awesome 👏 

I'll setup tickets for the follow up items we discussed

AnilSorathiya (approved):
well done!

PR #30, opened 2024-04-02
Improvements when using different models

cachafla:

- Use more informative `MissingOrInvalidModelPredictFnError` when `predict_proba` is not available on a model
  -  Side note: these tests should take probabilities via `assign_predictions()`
- Add new `UnsupportedModelForSHAPError` to properly inform the user that their model is not supported by our SHAP test. This can be fixed easily on our side, as long as you can actually compute SHAP on that model
- Support `shap.KernelExplainer` for SVC models
- Ensure we use `zero_division=0` when computing classification metrics so the annoying warning is not shown all the time

Tested a multitude of models. We can keep going to fully test `sklearn` but the fixes in this PR address support for most models:

```py
# RandomForestClassifier
from sklearn.ensemble import RandomForestClassifier

model = RandomForestClassifier()
model.fit(x_train, y_train)

# DecisionTreeClassifier
from sklearn.tree import DecisionTreeClassifier

model = DecisionTreeClassifier()
model.fit(x_train, y_train)

# LogisticRegression

model = LogisticRegression()
model.fit(x_train, y_train)

# SVC, LinearSVC
from sklearn.svm import SVC, LinearSVC

model = LinearSVC(dual=False)
# or
model = SVC()
model.fit(x_train, y_train)
```


johnwalz97 (approved):
Good stuff 👌

PR #41, opened 2024-04-11
Script to render notebooks as HTML with Quarto

cachafla:


nrichers (commented):
Nice demo PR! 🚀 This script shows what could be possible but might need some additional work to be ready for production. 

My main suggestion would be to turn this into a 2-step process:

1. **Use this script for a CI check**: Make sure notebooks execute without errors before  merging to prepare for step 2.
2. **Execute the notebooks when rendering the docs site**: I didn't realize this until I saw the output, but this script doesn't generate content we can overlay with the docs site as-is. The output is missing the docs site top navbar, sidebar, fonts, and extensive CSS theming. All of this could be fixed by executing the notebooks as part of the docs site build, which should be much safer if they have passed the CI check in step 1.

I also ran into some minor errors, cf. our Slack convo.

cachafla:
Closing for now. We can revisit later.

PR #43, opened 2024-04-15
[SC-3423] Review credit scorecard notebook

juanmleng:

- Add support for passing `prediction_probabilities` to `assign_predictions`: 

```
vm_train_ds.assign_predictions(
    model=vm_model,
    prediction_values=train_binary_predictions,
    prediction_probabilities = train_probabilities,
)
```
- Create separated demo `assign_prediction_probabilities.ipynb` to not overload assign predictions notebook.
- Remove logistic regression dedicated tests
- Make some changes in some tests to support `prediction_probabilities`


validmind/vm_models/dataset.py
```diff
@@ -423,17 +460,57 @@ def __assign_prediction_values(self, model, pred_column, prediction_values):
...
+            self._raw_dataset = np.hstack(
+                (self._raw_dataset, np.array(prediction_probabilities).reshape(-1, 1))
+            )
+            self._df[prob_column] = prediction_probabilities
+
+        # Update the dataset columns list
+        if prob_column not in self._columns:
+            self._columns.append(prob_column)
+
     def assign_predictions(  # noqa: C901 - we need to simplify this method
```
AnilSorathiya, line 491: This method is getting complex. We should optimize and/or split into multiple methods later.

validmind/vm_models/dataset.py
```diff
@@ -448,20 +525,67 @@ def assign_predictions(  # noqa: C901 - we need to simplify this method
...
                 raise ValueError(
                     "Length of prediction values doesn't match number of rows of the dataset"
                 )
             pred_column = f"{model.input_id}_prediction"
             if pred_column in self.columns:
-                raise ValueError(
-                    f"Prediction column {pred_column} already exists in the dataset"
+                warnings.warn(
+                    f"Prediction column {pred_column} already exists in the dataset, overwriting the existing predictions",
+                    UserWarning,
```
AnilSorathiya, line 539: Rather than overwriting the existing predictions, it would be nice to pass `overwrite_flag = True` as a method argument. It might happen that user forget to change `input_id` of the model.

validmind/vm_models/dataset.py
```diff
@@ -479,7 +603,46 @@ def assign_predictions(  # noqa: C901 - we need to simplify this method
             )
 
             prediction_values = np.array(model.predict(x_only))
-            self.__assign_prediction_values(model, pred_column, prediction_values)
+
+            # Check if the prediction values are probabilities
+            if _is_probability(prediction_values):
+
+                threshold = 0.5
```
AnilSorathiya, line 610: Why are we hard coding threshold here? It would be model developer's job to convert probabilities into `y_true`.

validmind/tests/model_validation/statsmodels/CumulativePredictionProbabilities.py
```diff
@@ -0,0 +1,140 @@
...
+    name = "cumulative_prediction_probabilities"
+    required_inputs = ["model", "datasets"]
+    metadata = {
+        "task_types": ["classification"],
+        "tags": ["logistic_regression", "visualization"],
+    }
+    default_params = {"title": "Cumulative Probabilities"}
+
+    @staticmethod
+    def plot_cumulative_prob(dataframes, dataset_titles, target_col, title):
```
AnilSorathiya, line 64: You can remove hardcoded `class_colors` using `matplotlib` for using  the colormap object. Then use these colors in the plotly lib:

```python
import numpy as np
import plotly.graph_objects as go
from matplotlib import cm

def plot_cumulative_prob(dataframes, dataset_titles, target_col, title):
    figures = []

    # Generate a colormap and convert to Plotly-accepted color format
    # Adjust 'viridis' to any other matplotlib colormap if desired
    colormap = cm.get_cmap('viridis')

    for i, (df, dataset_title) in enumerate(zip(dataframes, dataset_titles)):
        fig = go.Figure()

        # Get unique classes and assign colors
        classes = sorted(df[target_col].unique())
        colors = [colormap(i / len(classes))[:3] for i in range(len(classes))]  # RGB
        color_dict = {cls: f'rgb({int(rgb[0]*255)}, {int(rgb[1]*255)}, {int(rgb[2]*255)})' for cls, rgb in zip(classes, colors)}

        for class_value in classes:
            # Calculate cumulative distribution for the current class
            sorted_probs = np.sort(df[df[target_col] == class_value]["probabilities"])
            cumulative_probs = np.cumsum(sorted_probs) / np.sum(sorted_probs)

            fig.add_trace(
                go.Scatter(
                    x=sorted_probs,
                    y=cumulative_probs,
                    mode="lines",
                    name=f"{dataset_title} {target_col} = {class_value}",
                    line=dict(color=color_dict[class_value]),
                )
            )

        fig.update_layout(
            title_text=f"{title} - {dataset_title}",
            xaxis_title="Probability",
            yaxis_title="Cumulative Distribution",
            legend_title=target_col,
        )
        figures.append(fig)

    return figures

    ```

File: validmind/tests/model_validation/statsmodels/GINITable.py [hunk 5]
```diff
@@ -65,72 +65,50 @@ class GINITable(Metric):
...
 
         return self.cache_results(
             {
                 "metrics_summary": summary_metrics.to_dict(orient="records"),
             }
         )
 
-    def compute_metrics(self, model, X_train, y_train, X_test, y_test):
-        """Computes AUC, GINI, and KS for train and test sets."""
+    def compute_metrics(self):
```
AnilSorathiya, line 77: Can this test be converted into composite metrics later? may be not now.

validmind/tests/model_validation/statsmodels/PredictionProbabilitiesHistogram.py
```diff
@@ -68,95 +66,76 @@ class LogisticRegPredictionHistogram(Metric):
...
-        train_0 = df_train[df_train[target_col] == 0][pd_col]
-        train_1 = df_train[df_train[target_col] == 1][pd_col]
-        test_0 = df_test[df_test[target_col] == 0][pd_col]
-        test_1 = df_test[df_test[target_col] == 1][pd_col]
-
-        fig = make_subplots(rows=1, cols=2, subplot_titles=("Train Data", "Test Data"))
-
-        trace_train_0 = go.Histogram(
-            x=train_0, opacity=0.75, name=f"Train {target_col} = 0"
+    def plot_prob_histogram(dataframes, dataset_titles, target_col, title):
```
AnilSorathiya, line 69: The same here. we can use colormap to dynamically generate color of each class.

validmind/tests/model_validation/statsmodels/ScorecardHistogram.py
```diff
@@ -53,120 +49,88 @@ class ScorecardHistogram(Metric):
...
-
-        trace_train_0 = go.Histogram(
-            x=scores_train_0, opacity=0.75, name=f"Train {target_col} = 0"
-        )
-        trace_train_1 = go.Histogram(
-            x=scores_train_1, opacity=0.75, name=f"Train {target_col} = 1"
+    def plot_score_histogram(dataframes, dataset_titles, score_col, target_col, title):
+        figures = []
+        # Define a color map for multiple classes
+        class_colors = {
```
AnilSorathiya, line 66: use colormap to generate color dynamically

validmind/tests/model_validation/statsmodels/GINITable.py
```diff
@@ -65,72 +65,50 @@ class GINITable(Metric):
...
-            ["Train", "Test"], [X_train, X_test], [y_train, y_test]
-        ):
-            y_scores = model.predict(X)
+            # Retrieve y_true and y_pred for the current dataset
+            y_true = np.ravel(dataset.y)  # Flatten y_true to make it one-dimensional
+            y_prob = dataset.y_prob(self.inputs.model.input_id)
 
-            print("Predicted scores obtained...")
+            # Compute metrics
+            print(f"Computing metrics for {dataset_label}...")
```
cachafla, line 94: No need for this level of verbosity:

```suggestion
```

validmind/tests/model_validation/statsmodels/ScorecardHistogram.py
```diff
@@ -53,120 +49,88 @@ class ScorecardHistogram(Metric):
...
+            figures.append(fig)
+        return figures
 
     def run(self):
-        model = (
-            self.inputs.model[0]
-            if isinstance(self.inputs.model, list)
-            else self.inputs.model
+        title = self.params.get("title", self.default_params["title"])
+        score_column = self.params.get(
```
cachafla, line 109: This should already give you the value of `score_column` since a default value is defined:


```py
score_column = self.params["score_column"]
```

As a next step I suggest checking if the score column exists in the dataset and throw an error if it doesn't exist. It will catch users by surprise when they try to run the test without passing any parameters and see that we automatically try to access a `score` column.

validmind/vm_models/dataset.py
cachafla: I think in this case it's fine since the developer only passed a model and is letting us decide the threshold.

I would suggest we add a warning here that tells the user something like "using default threshold of 0.5 to assign class labels".

If the user has their own threshold they will need to pass prediction_values and prediction_probabilities manually.

cachafla: Nevermind, I see that you already do this here: https://github.com/validmind/developer-framework/pull/43/files#diff-19f766812f0e37e809d1eda669f0a45160d04616ba506156606dc1965cbcbce5R618

validmind/vm_models/dataset.py
```diff
@@ -479,7 +603,46 @@ def assign_predictions(  # noqa: C901 - we need to simplify this method
...
+                    logger.info("Running predict_proba()... This may take a while")
+                    prediction_probabilities = np.array(model.predict_proba(x_only))
+                    prob_column = f"{model.input_id}_probabilities"
+                    self.__assign_prediction_probabilities(
+                        model, prob_column, prediction_probabilities
+                    )
+                except MissingOrInvalidModelPredictFnError:
+                    # Log that predict_proba is not available or failed
+                    logger.warn(
+                        f"SKlearn model {model.__class__} Model does not have a compatible predict_proba implementation."
```
cachafla, line 641: Message has to be generic:

```suggestion
                        f"Model class '{model.__class__}' does not have a compatible predict_proba implementation."
```

validmind/tests/model_validation/statsmodels/GINITable.py
```diff
@@ -65,72 +65,50 @@ class GINITable(Metric):
...
 
-        for dataset, X, y in zip(
-            ["Train", "Test"], [X_train, X_test], [y_train, y_test]
-        ):
-            y_scores = model.predict(X)
+            # Retrieve y_true and y_pred for the current dataset
+            y_true = np.ravel(dataset.y)  # Flatten y_true to make it one-dimensional
+            y_prob = dataset.y_prob(self.inputs.model.input_id)
 
-            print("Predicted scores obtained...")
```
johnwalz97, line 98: One thing I'd recommend is to use `logger.info` and `logger.debug` versus `print` to print stuff out anywhere in library code. This gives us the opportunity to have different levels of verbosity. So if we are having issues we can optionally set `LOG_LEVEL=debug` to get more verbose output.

It also lets us control logging from a central config which can come in handy :)

johnwalz97 (approved):

cachafla (approved):
Great 👍 

For next steps I'd suggest completing the "TBC" documentation in the notebook so we can put it in the `code_samples/` directory and make it available to users.

We can do it with a follow up PR.

PR #44, opened 2024-04-16
John6797/sc 3882/metric decorator should work with test providers

johnwalz97:

This PR continues the effort to simplify metric creation. It adds support for functional metrics in the internal validmind test library as well as for external test providers.

Changes:
- add support for function types in the test loader
- rewrite the test providers notebook to show the new changes and just to improve it for better documentation
- add support to `save()` decorated metric functions
- change `y_pred()` method on the dataset class to take a `model` object instead of a `model_id` - this simplifies metric code and IMO makes a bit more sense as a user who's not familiar with input ids and how they work
- removing unit metric base class and changing all unit metrics to be pure functions - since we've just started adding them I think we should start off using functions instead of classes
- added support for unit metrics in the decorator/test provider/internal library by just checking if the return type of the function is a scalar value

## External Release Notes

Adding support for functional metrics for internal tests and test providers. Continuing our efforts to simplify the process for getting your custom code working in ValidMind, we now support functional metrics i.e. tests that can be defined as pure functions without any boilerplate.

See how this works and more in our [updated tutorial on custom test providers](https://docs.validmind.ai/notebooks/code_samples/custom_tests/external_test_providers.html)


notebooks/code_samples/custom_tests/external_test_providers.ipynb
```diff
@@ -1,329 +1,739 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Key Concepts\n",
+    "\n",
+    "- **Test Provider**: A Python class that gets registered with the ValidMind Framework and loads tests based on a test ID e.g. `my_test_provider.my_test_id`.\n",
+    "- **LocalTestProvider**: A built-in test provider that loads tests from a local directory.\n",
+    "- **Test ID**: A unique identifier for a test that is used to load the test from a test provider. Test IDs in ValidMind have multiple parts, separated by dots (`.`), that have special meaning/functions - `namespace.path.test_name:result_id` (e.g. `validmind.data_validation.ClassImbalance:1`):\n",
+    "    - `namesapce`: The namespace of the test - built-in tests use the `validmind` namespace, custom test providers specify a custom namespace when they are registered\n",
```
cachafla, line 35:
```suggestion
    "    - `namespace`: The namespace of the test - built-in tests use the `validmind` namespace, custom test providers specify a custom namespace when they are registered\n",
```

notebooks/code_samples/custom_tests/external_test_providers.ipynb
cachafla, line 1: In this notebook we only run metrics from the test provider. After running the previous notebook there will be two "providers" when add a new metric using the add test driven block functionality:

- "My Custom Metrics"
- "My Test Provider"

For this notebook I'd suggest updating the text and the screenshot at the end so we direct the user to add the blocks for "My Test Provider" instead. There are no guarantees that they have run the previous notebook on the same model.

notebooks/how_to/run_unit_metrics.ipynb
```diff
@@ -74,10 +74,10 @@
...
     "\n",
     "vm.init(\n",
-    "    api_host=\"https://api.prod.validmind.ai/api/v1/tracking\",\n",
-    "    api_key=\"...\",\n",
-    "    api_secret=\"...\",\n",
-    "    project=\"...\",\n",
+    "  api_host = \"https://api.dev.vm.validmind.ai/api/v1/tracking\",\n",
+    "  api_key = \"dea7a1c03b4bff0e4b969c9719d89492\",\n",
+    "  api_secret = \"65c3deb6f1c579a5e8a2aa3989c04068e77c42d5ab5e7c89ffd17aff25b291cc\",\n",
+    "  project = \"clukap3z500cd1ol2d8lgk6we\"\n",
```
cachafla, line 80:
```suggestion
    "  api_host = \"...\",\n",
    "  api_key = \"...\",\n",
    "  api_secret = \"...\",\n",
    "  project = \"...\"\n",
```

validmind/tests/__init__.py
```diff
@@ -404,7 +424,9 @@ def run_test(
     )
 
     test.run()
-    test.result.show()
+
+    if show:
+        test.result.show()
```
cachafla, line 429: Great

cachafla:
The new `save()` is a fancy trick 😄

validmind/ai.py
```diff
@@ -71,7 +71,7 @@ def __get_client_and_model():
 
     if "OPENAI_API_KEY" in os.environ:
         __client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
-        __model = os.environ.get("VM_OPENAI_MODEL", "gpt-4-turbo-preview")
```
juanmleng, line 74: We can address this later, but it might be worth considering giving users the option to choose their model, with GPT-4 as the default.

validmind/client.py
juanmleng, line 1: And polars as well?

juanmleng, line 1: `VMDataset` docstring mentions we only support `Pandas`. After adding `Polars`, does this still apply?

juanmleng (approved):
I left a couple of minor suggestions. Other than that, bravo!

notebooks/how_to/run_unit_metrics.ipynb
johnwalz97: ah good catch... thank you 🙏

notebooks/code_samples/custom_tests/external_test_providers.ipynb
johnwalz97: Yep thats a good point... Still have some thoughts on improving that ID problem between inline and provider tests so I might come back to that but for now I'll update :)

validmind/ai.py
johnwalz97: Yep thats a good idea... I think ideally we move to using our own API instead of calling the LLM directly from the dev framework but I am open to ideas there. This was more for a temporary solution AFAIK.

validmind/client.py
johnwalz97: good call... i will change that to show all of the new supported dataset types

PR #51, opened 2024-04-18
Fix prediction probabilities

juanmleng:

- fix bug in `is_probability` 
- fix bug in `prediction_values` when computed directly from model within validmind


validmind/vm_models/dataset.py
```diff
@@ -500,7 +500,9 @@ def _is_probability(output):
             """Check if the output from the predict method is probabilities."""
             # This is a simple check that assumes output is probabilities if they lie between 0 and 1
             if np.all((output >= 0) & (output <= 1)):
-                return True
+                # Check if there is at least one element that is neither 0 nor 1
+                if np.any((output > 0) & (output < 1)):
+                    return True
             return False
```
johnwalz97, line 506:
```suggestion
return np.all((output >= 0) & (output <= 1)) and np.any((output > 0) & (output < 1))
```

Should be able to simplify this into a single return :)

johnwalz97 (approved):
LGTM... just suggested a small change to simplify one of the functions

cachafla (approved):

PR #53, opened 2024-04-24
Hotfix for Integration Tests Script

johnwalz97:

The integration tests started failing after my latest merge. Turns out the code that replaces existing `vm.init` cells with a version that points to the project/endpoint for the integration tests was simply looking for `import validmind as vm` in each cell and replacing the whole cell if it had that import. This did not work for the new notebooks I added to the integration tests script so I updated to use `re.sub` to directly swap `vm.init` calls.


cachafla (approved):
👌

PR #55, opened 2024-04-25
use more robust markdown library

johnwalz97:

Switching to using `mistune` over the `markdown` library we were using. Mistune is quite a bit more robust and is way less strict about the markdown syntax. It also gives us the ability to add custom formatters etc in the future for stuff like mermaid diagrams, syntax highlighted code, math equations etc.


AnilSorathiya (approved):
👍

cachafla (commented):
Great 👏

PR #67, opened 2024-05-02
Support for external models

cachafla:

- Added unit test coverage to `Dataset` class
- Added support for model `attributes` when initializing a `Model`

## External Release Notes

You can now run documentation tests without passing a Python-native model object. This enables documenting:
- Models that are developed in non-Python environments
- Non-standard model interfaces:
    - Models deployed as APIs, i.e. SageMaker model endpoints
    - Tools such as Spark where a "model"  is not a typical object that exposes a `predict()` interface

To run tests for these models you typically need to load model predictions from a file, dataset, etc. The new `init_model` interface does not enforce a Python model object anymore. You only need to pass `attributes` that describe the model. This is required as a best practice for model documentation.

### Initializing an external model

Since there is no native Python object to pass to `init_model`, you will need to pass attributes that describe the model:

```python
# Assume you want to load predictions for a PySpark ML model
model_attributes = {
    "architecture": "Spark",
    "language": "PySpark",
}

# Or maybe you're loading predictions for a SageMaker endpoint (model API)
model_attributes = {
    "architecture": "SageMaker Model",
    "language": "Python",
}

# Call `init_model` without passing a model. Pass `attributes` instead
vm_model = vm.init_model(
    attributes=model_attributes,
    input_id="model",
)
```

### Assigning predictions

Since there's no model object available, the developer framework won't be able to call `model.predict()` (and `model.predict_proba()` for you. You will need to load predictions (and probabilities) manually, i.e.:

```python
vm_train_ds.assign_predictions(
    model=vm_model,
    prediction_values=prediction_values,
    prediction_probabilities=prediction_probabilities,
)
```

You can run tests as usual. No other changes are needed.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -108,17 +108,22 @@ class TestInitModel(TestCase):
         vm_model = init_model(model)
         self.assertEqual(vm_model.model, model)
 
-    @mock.patch(
-        "validmind.client.log_input",
-        return_value="1234",
-    )
-    def test_init_model_metadata_dict(self, mock_log_input):
+    def test_init_model_invalid_metadata_dict(self):
+        # Model metadata requires architecture and language at a minimum
+        metadata = {
+            "key": "value",
+            "foo": "bar",
+        }
+        with self.assertRaises(UnsupportedModelError) as context:
+            init_model(attributes=metadata, __log=False)
+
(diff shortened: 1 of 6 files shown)


AnilSorathiya (approved):
👏

PR #72, opened 2024-05-04
Fix rendering risk assessment blocks

cachafla:

With the fix for returning a validation report template when the user is a validator, we needed to make a small fix to properly render `Risk Assessment` blocks since it was throwing an error.

image


johnwalz97 (approved):
nice 😎

PR #77, opened 2024-05-09
Initial Framework Support for LLM Agents

johnwalz97:

This PR adds a lot of stuff related to GenAI use cases:

* Adds a new demo notebook that shows how you can use the developer framework with agent-like workflows where an LLM is using tools and function calls to retrieve information.
* Adds more data to the outputs of the ragas metrics
* Adds descriptions for the ragas metrics
* Adds support for latex in test/metric descriptions
* Adds the ability to configure complex input mappings for ragas metrics columns... We should look into a standard way of doing this since it could be useful for other things

Check out this notebook: https://github.com/validmind/developer-framework/blob/e63e9040bdb2dfb89495984b36f6507244ffa943/notebooks/code_sharing/llm/vendor_contract_agent/genai_vendor_contract_agent_usecase_poc.ipynb

## External Release Notes

**Initial support for agent-like workflows**: As part of an ongoing effort to improve the support for GenAI workflows in the ValidMind Developer Framework, we are adding support and examples/documentation for agent-like workflows. These are LLM "models" where the LLM can make a plan, collaborate with other LLMs to complete that plan, use tools and APIs to retrieve information from external systems etc.
**Support for LaTex in Test Descriptions**: You can now add LaTex formulas to any test description and it will be correctly rendered and displayed both in the developer framework as well as when logged to the ValidMind platform.


cachafla (approved):
Let's go 💪

PR #81, opened 2024-05-10
RAG Demo Notebook

johnwalz97:

Pulling in everything that we've been working on and adding some nice explanations and organization to the rag notebook... I also added support for function models and pipeline models in `init_model`


juanmleng (approved):
Looks awesome :)

PR #92, opened 2024-05-17
Allow passing section_id and position to result.log()

cachafla:
### New Features for `result.log()`

New enhancements have been added to the logging of individual results with the Developer Framework. Recall that individual tests can be run with `run_test` and this function returns a `result` object that allows logging to the results to the ValidMind API with a call to `.log()`:

```python
result = vm.tests.run_test(
    "validmind.data_validation.ClassImbalance",
    inputs={"dataset": vm_dataset},
)

result.log()
```

The following features have been added to the `.log()` function:

1. **Specify a Section**: 
   - You can now log the result to a specific section by using the `section_id` parameter.
   - Example:
     ```python
     result.log(section_id="model_overview")
     ```

2. **Specify a Section and Position**:
   - You can also specify the position within the section where the result should be logged. Positions start at 0.
   - Invalid positions will throw an error.
   - Example:
     ```python
     result.log(section_id="model_overview", position=2)
     ```

3. **Prevent Duplicate Logs**:
   - Calling the `.log()` method repeatedly for the same result will not create multiple blocks in the document. It ensures only one block is added at the specified location.
   - Example:
     ```python
     result.log(section_id="model_overview")
     result.log(section_id="model_overview")
     result.log(section_id="model_overview")
     ```
   - This sequence will only add one block at the end.

These enhancements make it easier to organize and manage your test result logs, providing a more structured and efficient logging process.

### DIFF
--- a/validmind/errors.py
+++ b/validmind/errors.py
@@ -339,7 +339,7 @@ def raise_api_error(error_string):
     try:
         json_response = json.loads(error_string)
         api_code = json_response.get("code")
-        api_description = json_response.get("description")
+        api_description = json_response.get("description", json_response.get("message"))
     except json.decoder.JSONDecodeError:
         api_code = "unknown"
         api_description = error_string
--- a/tests/test_api_client.py
+++ b/tests/test_api_client.py
@@ -239,8 +239,8 @@ class TestAPIClient(unittest.TestCase):
 
         mock_log_test_result.assert_has_calls(
             [
-                call(results[0], ["input1"], "training"),
-                call(results[1], ["input1"], "training"),
+                call(results[0], ["input1"]),
+                call(results[1], ["input1"]),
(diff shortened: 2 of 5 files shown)


validmind/errors.py
```diff
@@ -339,7 +339,7 @@ def raise_api_error(error_string):
     try:
         json_response = json.loads(error_string)
         api_code = json_response.get("code")
-        api_description = json_response.get("description")
+        api_description = json_response.get("description", json_response.get("message"))
```
cachafla, line 342: This is the reason all those API errors were so cryptic, we were not accessing the correct element.

johnwalz97: oops 🙈

johnwalz97 (approved):
Awesome 😎

PR #102, opened 2024-05-25
Add support for the ValidMind LLM Service

johnwalz97:

This PR adds support for the new LiteLLM proxy that we will deploying as a new service on the platform. This service gives us the ability to proxy OpenAI api requests to OpenAI or any other LLM while using a master API key that is stored on the platform and virtual keys for individual users. It also lets us manage budgets, rate limiting and much more on a per-user or per-organization level. This means that we can offer an out-of-the box LLM API as an addon or free feature without having to issue OpenAI API keys to individual users. It also means we can use the same `openai` library on the dev framework side while proxying to any closed or open-source LLM behind the scenes. See this PR: https://github.com/validmind/backend/pull/572 for more info on how this is implemented.

### Implementation

For now, how this will work is that whenever code in the dev framework needs to use an LLM, the following logic will be run:

1. First, the dev framework will look for an `OPENAI_API_KEY` or a `AZURE_OPENAI_KEY` in the environment variables.
2. If either is found, it will use those credentials to connect directly to the appropriate API.
3. If neither environment variable is set, it will call the ValidMind API (configured during `vm.init()`) to attempt to retrieve a virtual API key and the URL for the ValidMind LLM proxy.
4. If a 200 response is returned along with the key and url, then the framework will use the ValidMind LLM configured by the proxy.
5. If not, then an error will be thrown telling the user that they cannot use LLM features unless they configure their own via env vars or have a ValidMind account that includes LLM access.

Some more work will definitely be needed here in the future when we start implementing stuff like rate limiting and per-user or per-org LLM budgets. But, for now, this will allow us to start offering out-of-the-box LLM features that get proxied through our platform.

### Testing

1. To test this, first pull the latest changes from the backend repo (or switch to this branch if it hasn't been merged yet https://github.com/validmind/backend/pull/572).
2. Then start up the backend as usual but run the following in a separate terminal (or if you are running everything with docker, just run `docker compose up --profile all`

    ```bash
    docker compose up litellm
    ```
2. Then using this branch, just run any test or test suite in the dev framework but make sure to remove any OpenAI or Azure OpenAI keys from your env vars with this snippet at the top of the notebook:

    ```python
    import os
    del os.environ["OPENAI_API_KEY"]
    del os.environ["AZURE_OPENAI_KEY"]
    ```

## External Release Notes

Adding support for ValidMind LLM to use for LLM-as-Judge tests, prompt evaluation tests as well as test description and interpretation generation.


cachafla (approved):
Awesome 🫡

PR #103, opened 2024-05-29
Fix misspellings of "ValidMind"

noosheenv:

There were some cases where the proper noun form of "ValidMind" was not being followed, i.e. "Validmind". 

There were two cases where I didn't change it yet:
- In `run_unit_metrics.ipynb` because the text references a screenshot with the same misspelling:
image

- In `LanguageDetection.py` because it is part of an error message script and I'm not sure I should mess with that.
image


cachafla (approved):
Nice 👌

cachafla:
@noosheenv @nrichers OK to merge this?

noosheenv:
Yes! I think I got scared. Merging now

PR #105, opened 2024-06-04
run_test() integration tests

cachafla:

Integration tests run automatically with `make test` and the automated GitHub check `Python application / build (pull_request)`.

To just run the integration tests:

```
make test ONLY="tests.test_run_all_tests" 
```

To run tests just for a given set of test IDs (1 or more, comma separated):

```
TEST_IDS=validmind.data_validation.BivariateScatterPlots,validmind.data_validation.PiTCreditScoresHistogram make test ONLY="tests.test_run_all_tests"
```

### What this does

For every test in the `validmind/tests/` directory:

- Ensures it has `required_inputs`
- Ensures it has `metadata`
  - It has `task_types`
  - It has `tags`
- Tries to load test inputs (datasets, models) appropriate to the test type
- Tries to load specific config if available
- Calls `vm.tests.run_test()` with the required params
- Asserts the test returns a `MetricResultWrapper` or a `ThresholdTestResultWrapper`
- Asserts the return value contains at least a result object or figures

It also generates a report like below, noting:

- `SUCCESS`: the test ran successfully
- `KNOWN FAILURE`: the test is expected to fail so it didn't run
- `SKIPPED`: the test is skipped because we can't run it yet (need to support config/inputs). This is the caes for most LLM/embedding tests, for example

image


tests/test_run_all_tests.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+class TestRunTest(unittest.TestCase):
+    pass
+
+
+def create_unit_test_func(vm_test_id, vm_test_class):
+    def unit_test_func(self):
+        self.assertTrue(
+            hasattr(vm_test_class, "required_inputs"),
```
AnilSorathiya, line 95: The tests that are implemented using decorator functionality won't have `required_inputs` attributes. How do we handle these?

johnwalz97: The required inputs for function tests should get those set automatically when they are loaded via `load_test`... It basically just checks if the function arguments contains an input name (model, models, dataset, datasets) and sets those as required inputs if so.

johnwalz97 (approved):
Very nice 👌 This is going to be super helpful.

tests/test_run_all_tests.py
```diff
@@ -0,0 +1,300 @@
...
+        # Build the single test inputs according to the required inputs
+        single_test_inputs = {}
+        if required_inputs == ["dataset"]:
+            single_test_inputs = inputs["single_dataset"]
+        elif required_inputs == ["dataset", "model"]:
+            single_test_inputs = inputs["model_and_dataset"]
+        elif required_inputs == ["datasets"]:
+            single_test_inputs = inputs["two_datasets"]
+        elif required_inputs == ["datasets", "model"]:
+            single_test_inputs = inputs["model_and_two_datasets"]
```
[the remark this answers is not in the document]
cachafla: We can support those by adding an extra combination of inputs. And we can name those input combinations like we do for `model_and_two_datasets`. This file allows for that flexibility 👍

PR #106, opened 2024-06-07
Terminology consolidation: use "Test" to refer to a Metric or Threshold Test

cachafla:

- Rename `@vm.metric` to `@vm.tets`
- Stop exposing `Metric` and `ThresholdTest` when listing tests
- Consolidate `Test` terminology and just refer to `Tests` in end user notebooks. 

## External Release Notes

We have an overload of concepts on the developer framework since we use 3 words to describe roughly the same concept:

- Test: generic concept that maps test results to a test block in a document
- Metric: a "test" that generates some results
- Threshold Test: a "test" that can optionally take a threshold as input and generate a output wit pass/fail criteria

This has generally created confusion and made difficult for users to understand what to use and when.

The new decorator based tests (that use a function instead of a class) already hide this complexity so we won't need to expose the internal `Metric` and `ThresholdTest` classes any more.

In this PR we consolidate this terminology and just refer to `Tests` in our end user notebooks.


cachafla:
@nrichers @robinzimmermann examples of the how we have renamed `Metric` -> `Test` in notebooks:

image

image

johnwalz97 (approved):
Very nice 🙌

PR #112, opened 2024-06-11
[Low Priority] Add literal type for Test IDs

johnwalz97:

Quality of life feature:

Adds a script that auto runs whenever a commit is made that will generate and save a list of test IDs to `validmind/tests/__types__.py`. This lets us use a literal string type in our type hints wherever we accept a test ID as a param and then when you go to run that function in a notebook it will show a list of possible values. This doesn't work in Jupyter AFAIK but its quite helpful for us who are working in VS Code notebooks and want to be able to quickly do something like:

```python

result = run_test("validmind.really.cool.test", input={})
```

image

Please make sure to run `make install` to get the updated pre-commit hooks installed properly.


cachafla (approved):
Nice 👌

PR #114, opened 2024-06-12
FIxed the broken/incorrect anchors in the ToC for "Introduction for Model Developers"

validbeck:
| Old | New* |
|---|---|
|Screenshot 2024-06-12 at 8 59 17 AM|Screenshot 2024-06-12 at 8 55 30 AM|
| "Run `run_documentation_tests()` using `vm_raw_dataset_new` as input" | Changed to "Run `run_documentation_tests()` using `vm_raw_dataset_preprocessed` as input" |
| Log the individual result high correlation test using `vm_raw_dataset` (no data cleanup) | Changed to "Log the individual result of the high correlation test that used `vm_balanced_raw_dataset` (that had a highly correlated `Age` column) as input" |
| Log the individual result high correlation test using `vm_raw_dataset_new` (balanced classes and reduced features) | This section was removed, so I removed the anchor. |

*Styling is just the theme of my VSCode, not applied to the actual notebook.

### Testing
I tested all links in the ToC in the following environments/formats:

| Format | Outcome |
|---|---|
| `.ipynb Rich` Text Display in VSCode | Works as expected. |
| Quarto Preview HTML Output | Works as expected. |
| JupyterHub | Works as expected. **I uploaded my own edits to JupyterHub**, as I don't even see this notebook in the `ValidMind Demo` notebooks folder in Drive. I did check the uploaded version in `jupyterhub.validmind.ai` and do see some strange behaviour, but the edits should fix that! |
| Google Colab | **Our pre-generated ToC doesn't work at all, but their built in side-bar ToC does.** Seems to be expected behaviour when I quickly Googled it as [ToCs seem to be a different kind of cell gated to Colab Pro users](https://stackoverflow.com/questions/67458990/how-to-automatically-generate-a-table-of-contents-in-colab-notebook). |

### DIFF
--- a/notebooks/tutorials/intro_for_model_developers.ipynb
+++ b/notebooks/tutorials/intro_for_model_developers.ipynb
@@ -55,9 +55,8 @@
     "  - [Run some tabular data tests](#toc5_2_)\n",
     "  - [Utilize test output](#toc5_3_)\n",
     "  - [Documenting the results based on two datasets](#toc5_4_)\n",
-    "    - [Run `run_documentation_tests()` using `vm_raw_dataset_new` as input](#toc5_4_1_)\n",
-    "    - [Log the individual result high correlation test using `vm_raw_dataset` (no data cleanup)](#toc5_4_2_)\n",
-    "    - [Log the individual result high correlation test using `vm_raw_dataset_new` (balanced classes and reduced features)](#toc5_4_3_)\n",
+    "    - [Run `run_documentation_tests()` using `vm_raw_dataset_preprocessed` as input](#toc5_4_1_)\n",
+    "    - [Log the individual result of the high correlation test that used `vm_balanced_raw_dataset` (that had a highly correlated `Age` column) as input](#toc5_4_2_)\n",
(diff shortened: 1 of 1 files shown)


nrichers:
> JupyterHub | Works as expected. I uploaded my own edits to JupyterHub, as I don't even see this notebook in the `ValidMind Demo` notebooks folder in Drive

That's great news! 

FYI, only Colab sources the notebook out of Drive, JupyterHub sources them out of 

https://github.com/validmind/demo-environment/tree/main/demo_files

 which in turn is sourced out of 

https://github.com/validmind/developer-framework/tree/main/notebooks

nrichers (approved):
LGTM! Thank you so much for investigating and fixing. Much, much appreciated. 🙇

validbeck:
I just took a closer look and while the actual issues are resolved locally and in *JupyterLite* (the newer version of JupyterHub) I think the ToC extension is just broken for JupyterHub. I will do some investigation as to if we can fix this or if we need to upgrade our JupyterHub instance somehow...

PR #117, opened 2024-06-13
Chore: Dependency Updates

johnwalz97:

This PR cleans up our dependencies since we have added quite a few over the last few months and it's put a real strain on the dependency resolver and has far too strict version constraints. To fix this, I went through and relaxed the constraints on most of our dependencies. I also removed some unused dependencies, moved the ragas dependency into the `llm` extra and removed a few unnecessary constraints. This should make our package a lot more compatible with users' environments and make poetry's job a bit easier when it comes to updating and adding new dependencies.


cachafla (approved):
Much needed 😄

AnilSorathiya (approved):
Thanks! 👍

PR #123, opened 2024-06-18
Support for comparison tests

cachafla:
## Updated Description (from @johnwalz97):

After taking over this ticket from @cachafla and talking with the team, I have gone ahead and implemented the comparison test as a separate method which is called by `run_test()` whenever `input_grid` is passed as an argument. 

`input_grid` can be a dictionary where each key is the name of an input and the value is a list of different values for that input name. This will be expanded into a list of dictionaries that is the cartesian product (all possible pairwise combinations). `input_grid` can also be set to a list of dictionaries to have full control over the input groups or combinations to be run.

Once the `input_grid` is expanded, each combination is run as a normal test sequentially with `run_test`. Then the results are aggregated into a single result object. Summary tables are aggregated and a column is added for each input to indicate which input combination was used to produce each table row. Plots are simply all aggregated into a single list but each plot will be annotated with a description of the input combination used to produce them.

As part of this work I chose to refactor the `validmind.tests` module since the main module file was getting huge. I split out the single `__init__.py` file into different sub-modules that handle the different responsibilities that the `validmind.tests` module has. I also added the concept of a store to simplify the complexity of caching loaded tests. 

Finally, I refactored the base test class to move the `task_types` and `tags` properties out of the `metadata` object and set them as first-class properties of the test class (renamed `task_types` to simply `tasks`). This made the code that actually consumes these properties much simpler.

## Initial Description (from @cachafla):

Starting with `Option 2: Parameter Grid` here: https://chatgpt.com/share/e/314f8bdf-2837-4c2d-ac71-179e9189a933

Ideally the actual combination of results is unrelated to how we provide input combinations, so we can continue to refine what's the best developer interface.

Currently, this PR adds a new `run_comparison_test` function that allows running existing test in "comparison" mode, whether we want to compare performance of datasets, models or all of them.

To decide on the team (cc @AnilSorathiya @johnwalz97 @juanmleng):
- Should we build this interface into `run_test()` itself?
- Should we keep it separate like `run_comparison_test()` already does?

## External Release Notes

**Support for Comparison Tests**: A crucial feature has been added to the Developer Framework to support running comparison tests. This is useful for any sort of case where you might want to run the same test against multiple combinations of models or dataset and create a single documentation block that compares the individual results.

The updated `run_test()` function now allows you to pass an `input_grid` that will run a test for all combinations of inputs.

Example - for this input grid:

```python
input_grid = {
    "model": ["XGBoost"],
    "dataset": ["train_dataset", "test_dataset"],
}
```

A test will run once for each of the following input groups:

```python
{"model": "XGBoost", "dataset": "train_dataset"}
{"model": "XGBoost", "dataset": "test_dataset"}
```

### Example function calls

```python

input_grid = {
    "model": ["XGBoost"],
    "dataset": ["train_dataset", "test_dataset"],
}

result = run_test(
    "validmind.model_validation.sklearn.ClassifierPerformance",
    input_grid,
)

result = run_test(
    "validmind.model_validation.sklearn.ConfusionMatrix",
    input_grid,
)

result = run_test(
    "validmind.model_validation.sklearn.ROCCurve",
    input_grid,
)
```

### Demo video

https://github.com/validmind/developer-framework/assets/21595/af0bfbbf-dbc7-461c-ae7e-4e4ccb26544b


AnilSorathiya:
I like the idea to handle comparison tests separately using the interface the `run_comparison_test` but what if we don't expose to users. We use the `run_test` interface by adding the `input_grid` argument. The `run_test` method can  invoke the `run_comparison_test` if the `input_grid` is not `None`.  
One less interface for the customer.

juanmleng:
Regarding the `input_grid`, there are scenarios where using all pairwise combinations of models and datasets may not be convenient. For example, when different models are designed to be trained on specific datasets, we may require a custom input grid. Consider a case where we have two models, `XGBoost_1` and `XGBoost_2`, each trained on different datasets, `train_dataset_1` and `train_dataset_2`, respectively. The expected grid would look like this:
```
{"model": "XGBoost_1", "dataset": "train_dataset_1"}
{"model": "XGBoost_2", "dataset": "train_dataset_2"}
```
Additionally, in scenarios involving a high number of models and datasets, users might only be interested in documenting a subset of the possible pairwise combinations. For example, `XGBoost_v1` is tested with `test_dataset_A` and `test_dataset_B` because prior experiments on training data suggest these pairs show best results. And for the same reason, perhaps users only want to show `Random_Forest` paired only with `dataset_A`.

To support these use cases, a suggestion would be to allow users to define a custom input grid. As a caveat, this introduces an intermediate step for input grid setup, but gives more flexibility and control. It may be good idea to implement both, and allow users to either pass the current model and dataset lists or a pre-defined custom input grid.

johnwalz97:
I'm a big fan of what @AnilSorathiya said. We have a separate module for comparison tests but its exposed via `run_test()`. This will keep our current state of only having two entry points for running tests i.e. `run_test()` and `run_documentation_tests()`.

johnwalz97:
@juanmleng thats a good point... Maybe we have two different ways of passing inputs via `input_grid`. Either a dictionary like in the examples @cachafla shared. Or a list of dictionaries in which case you would have full control over the grid like in your example.

johnwalz97 (approved):
Approving on behalf of @cachafla since he created the PR initially and GH won't let PR authors approve their own PRs

PR #133, opened 2024-07-04
Fixed broken links referring to nlp_and_llm instead of NLP_and_LLM

noosheenv:
There were two references to `notebooks/code_samples/nlp_and_llm/foundation_models_integration_demo.ipynb` that were broken because the folder is named `NLP_and_LLM`. 

Changed the references to this: `notebooks/code_samples/NLP_and_LLM/foundation_models_integration_demo.ipynb` and the links work now!! Woohoo!

The notebooks affected are: 
`notebooks/how_to/use_dataset_model_objects.ipynb`
`notebooks/tutorials/intro_for_model_developers.ipynb`

For reference, both links were in this section of the notebooks (LLM model documentation demo):
image

### DIFF
--- a/notebooks/how_to/use_dataset_model_objects.ipynb
+++ b/notebooks/how_to/use_dataset_model_objects.ipynb
@@ -849,7 +849,7 @@
     "\n",
     "- [Application scorecard demo](../code_samples/credit_risk/application_scorecard_demo.ipynb)\n",
     "- [Linear regression documentation demo](../code_samples/regression/quickstart_regression_full_suite.ipynb)\n",
-    "- [LLM model documentation demo](../code_samples/nlp_and_llm/foundation_models_integration_demo.ipynb)\n",
+    "- [LLM model documentation demo](../code_samples/NLP_and_LLM/foundation_models_integration_demo.ipynb)\n",
     "\n",
     "<a id='toc7_2_'></a>\n",
     "\n",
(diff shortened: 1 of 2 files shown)


cachafla:
This is strange: does your `developer-framework` have those directories in capital letters? It should be lower case:

image

noosheenv:
> This is strange: does your `developer-framework` have those directories in capital letters? It should be lower case:
> 
> image

It turns out the link was broken on the docs site because we named our folder as `NLP_and_LLM` for reasons, but it's lowercase everywhere else. So if we fix it in one spot, it'll break on the other spot :( 

A happy solution would be to rename the folder in `developer-framework` to `NLP_and_LLM` so we match. Is it possible to do that on this PR please?

cachafla:
> > This is strange: does your `developer-framework` have those directories in capital letters? It should be lower case:
> > image
> 
> It turns out the link was broken on the docs site because we named our folder as `NLP_and_LLM` for reasons, but it's lowercase everywhere else. So if we fix it in one spot, it'll break on the other spot :(
> 
> A happy solution would be to rename the folder in `developer-framework` to `NLP_and_LLM` so we match. Is it possible to do that on this PR please?

I see. In that case it's better to fix it wherever the folder is being used. I wouldn't advise changing the developer framework just to adapt to changes in the doc generation pipeline.

noosheenv:
We figured out that these links were fine here, we just need to find a better workaround in the `documentation` repo. Closing!!

PR #149, opened 2024-07-22
Fix `Text`  type casting

cachafla:

Ensure we cast `Text` as a categorial field when generating field statistics.


johnwalz97 (approved):
👍

PR #152, opened 2024-07-29
[SC 5587] Initial support ongoing monitoring through header

AnilSorathiya:
We would like to support new content type, monitoring for tests so that we can differentiate between document tests and monitoring tests. These tests can be shown in the monitoring template.

 - Pass a `monitoring` parameter in the `vm.init()` method so that tests results to go to a different "bucket". By default value is `False`
 - Tests run after that should go into in `monitoring` bucket that will be handled in the backend if the `X-MONITORING`  header  is `True`.

```python
import validmind as vm
vm.init(
  api_host = "....",
  api_key = "....",
  api_secret = "....",
  project = "...",
  monitoring=True
)
```
## External Notes for Reviewers
 - Pass a `monitoring` parameter in the `vm.init()` method so that tests results to go to a different "bucket". By default value is `False`
 - Tests run after that should go into in `monitoring` bucket that will be handled in the backend if the `X-MONITORING`  header  is `True`.
 
```python
import validmind as vm
vm.init(
  api_host = "....",
  api_key = "....",
  api_secret = "....",
  project = "...",
  monitoring=True
)
```

### DIFF
--- a/tests/test_api_client.py
+++ b/tests/test_api_client.py
@@ -72,6 +72,7 @@ class TestAPIClient(unittest.TestCase):
                 "X-API-KEY": os.environ["VM_API_KEY"],
                 "X-API-SECRET": os.environ["VM_API_SECRET"],
                 "X-PROJECT-CUID": os.environ["VM_API_MODEL"],
+                "X-MONITORING": "False",
             },
         )
 
@@ -135,6 +136,8 @@ class TestAPIClient(unittest.TestCase):
                 "X-API-KEY": os.environ["VM_API_KEY"],
                 "X-API-SECRET": os.environ["VM_API_SECRET"],
                 "X-PROJECT-CUID": os.environ["VM_API_MODEL"],
+                "X-MONITORING": "False",
+
             },
         )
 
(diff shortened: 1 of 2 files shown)


tests/test_api_client.py
```diff
@@ -72,6 +72,7 @@ def test_init_successful(self, mock_requests_get):
                 "X-API-KEY": os.environ["VM_API_KEY"],
                 "X-API-SECRET": os.environ["VM_API_SECRET"],
                 "X-PROJECT-CUID": os.environ["VM_API_MODEL"],
+                "X-MONITORING": "False",
```
cachafla, line 75: I think it should be fine to use a boolean. The HTTP client will convert it to a string for us. We just need to ensure we parse the incoming values properly on the backend side.


```suggestion
                "X-MONITORING": False,
```

cachafla:
Instructions should use booleans:

```
import validmind as vm

vm.init(
  api_host = "....",
  api_key = "....",
  api_secret = "....",
  project = "...",
  monitoring=True
)
```

johnwalz97 (approved):
Good stuff! This will be a great way to support monitoring tests until we have a more proper "production" solution 👏👌

tests/test_api_client.py
AnilSorathiya: HTTP client doesn't covert the header into string if we pass it boolean.
```
InvalidHeader: Header part (True) from ('X-MONITORING', True) must be of type str or bytes, not <class 'bool'>
```
solution:
we can pass `vm.init(monitoring=True)` then covert in to str when passing to http client.

PR #155, opened 2024-08-06
Feat: Mechanism for Filtering Columns in a Dataset when Running a Test

johnwalz97:
### Background

This PR adds a generic mechanism for creating temporary versions of input (`VMDataset` or `VMModel`) objects with special options applied when running tests. This is done via a new `with_options()` method that is set but not implemented on the new `VMInput` base class and overridden on the `VMDataset` class only (for now). This method on the dataset class supports an option called `columns` that allows passing a list of column names to include in the new copy of the dataset that will get returned. Users will not call this method directly but instead can pass a dictionary with the following structure:
```python
{
    "input_id": "my_dataset",
    "columns": ["col1", "col2"],
    ... # other options can be supported in the future
}
```
instead of an input ID or input object when running a test.

### Example Usage

> see [new notebook in this PR](https://github.com/validmind/developer-framework/blob/f322a09eb9157e210408ab503cef6aa9a62197ee/notebooks/how_to/input_column_filtering.ipynb) for full example)

```python

result = run_test(
    test_id="validmind.data_validation.DatasetDescription",
    inputs={"dataset": {"input_id": "my_dataset", "columns": ["col1"]}},
)
```

### Changes

* Add new base class (`VMInput`) for input objects (`VMDataset` and `VMModel` objects) in `validmind/vm_models/input.py`... This base class for now just adds a `with_options()` method to be overridden by subclasses.
* Add a new `with_options()` method to the `VMDataset` class in `validmind/vm_models/dataset/dataset.py` that supports only a `columns` option for now that will return a filtered version of the dataset with only the specified columns included.
* Add support for passing dictionaries as inputs in the `TestContext` class in `validmind/vm_models/test_context.py` so that users can pass an input dictionary where the key `input_id` points to an input object in the input registry and the rest of the keys act as options that will be passed to the `with_options()` method of the input object. The new copy of the input object returned by `with_options()` will be passed to the test when its run.
* Fix an issue introduced by my last [PR](https://github.com/validmind/developer-framework/pull/151) where one of the integration test notebooks (`notebooks/how_to/run_unit_metrics.ipynb`) fails when fingerprinting the input dataset for a unit metric due to using `dataset.df` instead of `dataset._df`.

## External Release Notes

**Support for Column Filtering when Running Tests**:

Tests can now be run on a subset of columns in a dataset without having to create a whole new dataset with the columns removed. See the following example that shows how to run the Dataset Description test with only a single column from a dataset:
```python

result = run_test(
    test_id="validmind.data_validation.DatasetDescription",
    inputs={"dataset": {"input_id": "my_dataset", "columns": ["col1"]}},
)
# notice how the `dataset` input is set to a dictionary whose key `input_id` maps to a particular input dataset
# and whose `columns` key sets the subset of columns from the original dataset to pass to the Dataset Description test
```

See this [example notebook](https://github.com/validmind/developer-framework/blob/main/notebooks/how_to/input_column_filtering.ipynb) for more information.


notebooks/how_to/input_column_filtering.ipynb
cachafla, line 1: Small: I suggest renaming the notebook so that it can be read along with `how_to`:

`notebooks/how_to/filter_input_columns.ipynb`

cachafla (approved):
Awesome 👏

PR #158, opened 2024-08-07
[SC-5732] Fix inconsistent AUC calculations using class predictions instead of prediction probabilities

juanmleng:
### Description
Some of our classification performance tests compute ROC AUC based on class predictions, while others use prediction probabilities. The correct way to compute ROC AUC is by using prediction probabilities. Currently, the tests `ClassifierPerformance` and `MinimumROCAUCScore` and unit metric `ROC_AUC` compute AUC using class predictions, whereas `ROCCurve` uses prediction probabilities. This results in different AUC values for each of these tests. This PR aims to fix this issue by ensuring that all tests using ROC AUC use the same method—prediction probabilities.

### Test
- Using the dataset and model from `quickstart_customer_churn_full_suite.ipynb`, running the test `MinimumROCAUCScore`, `ClassifierPerformance` and `ROCCurve` should produce the same AUC values. 
- Successfully run the `run_unit_metrics.ipynb` notebook, and the value of the unit metric `ROC_AUC` should be higher after this change.


validmind/tests/model_validation/sklearn/ClassifierPerformance.py
```diff
@@ -126,9 +126,18 @@ def run(self):
...
             zero_division=0,
         )
-        report["roc_auc"] = multiclass_roc_auc_score(
-            self.inputs.dataset.y,
-            self.inputs.dataset.y_pred(self.inputs.model),
-        )
+
+        y_true = self.inputs.dataset.y
+
+        if len(unique(y_true)) > 2:
```
johnwalz97, line 132:
```suggestion
        if len(np.unique(y_true)) > 2:
```

I think using `np.unique` here instead of `unique` might be a bit more performant especially for larger datasets but I am not 100% sure on that?

A better long-term solution might be to check the number target classes set on the dataset class: `len(self.inputs.dataset.target_class_labels)`. The problem there is its not guaranteed that the user passes the `class_labels` argument when running `init_dataset` so it could be wrong. I am thinking we update this line of the dataset class: https://github.com/validmind/developer-framework/blob/main/validmind/vm_models/dataset/dataset.py#L108 to be the following:

```python
        # if class labels not supplied, use unique values in target column
        self.target_class_labels = (
            target_class_labels
            if target_class_labels
            else {str(i): str(i) for i in np.unique(self._df[self.target_column])}
        )
```

What do you think?

validmind/tests/model_validation/sklearn/MinimumROCAUCScore.py
```diff
@@ -99,9 +100,15 @@ def multiclass_roc_auc_score(self, y_test, y_pred, average="macro"):
 
     def run(self):
         y_true = self.inputs.dataset.y
-        class_pred = self.inputs.dataset.y_pred(self.inputs.model)
-        y_true = y_true.astype(class_pred.dtype)
-        roc_auc = self.multiclass_roc_auc_score(y_true, class_pred)
+
+        if len(unique(y_true)) > 2:
```
johnwalz97, line 104:
```suggestion
        if len(np.unique(y_true)) > 2:
```

validmind/unit_metrics/classification/sklearn/ROC_AUC.py
```diff
@@ -3,11 +3,32 @@
...
 
 
 @tags("classification", "sklearn", "unit_metric")
 @tasks("classification")
 def ROC_AUC(model, dataset, **kwargs):
-    return roc_auc_score(dataset.y, dataset.y_pred(model), **kwargs)
+
+    y_true = dataset.y
+
+    if len(unique(y_true)) > 2:
```
johnwalz97, line 18:
```suggestion
    if len(np.unique(y_true)) > 2:
```

johnwalz97 (approved):
Looks good minus some nit picks 👍

validmind/tests/model_validation/sklearn/ClassifierPerformance.py
cachafla: Yep, `np.unique(y_true)` works since we can't guarantee the user will pass `target_class_labels`

cachafla (commented):
Looks good. Should be ready to go after `np` is used to compute unique values.

validmind/tests/model_validation/sklearn/ClassifierPerformance.py
juanmleng: I like the idea of setting and using the `target_class_labels`. I will use `np.unique(y_true)` for this one, but happy to introduce the idea of setting `target_class_labels` going forward.

PR #161, opened 2024-08-08
Fixes to multiple tests

cachafla:

Changes:

- PrecisionRecallCurve uses probabilities
- The `OverfitDiagnosis` test now handles both classification and regression models and uses auc for classification and mse for regression by default. Support for user-configurable performance metrics has also been added (one of `accuracy`, `auc`, `f1`, `precision`, `recall`, `mse`, `mae`, `r2`, or `mape` depeding on the model task)
- The `RobustnessDiagnosis` test has been updated to use the ROC_AUC score instead of accuracy for evaluating model robustness


johnwalz97 (approved):
LGTM... Approving since @cachafla opened this PR originally

PR #163, opened 2024-08-09
Update shap to 0.44.1

cachafla:

Current `shap` version was throwing the following error on JupyterHub:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbe in position 370: invalid start byte
```

https://github.com/shap/shap/issues/1215


cachafla:
Approved.

johnwalz97 (commented):
Approving after the fact ✅

PR #166, opened 2024-08-12
[SC-5784] Fix errors in unit metrics notebook

juanmleng:
- Updated the `run_unit_metrics.ipynb` notebook to use a supported ValidMind model documentation project
- Replaced screenshots to showcase the updated add a text-driven block functionality using checkboxes 
- Fixed a small bug where `test_id` was using `composite_test` instead of `composite_metric`


cachafla (approved):
👌

PR #176, opened 2024-08-20
First-Class Support for Unit Metrics (Phase 1)

johnwalz97:
Shortcut Story: https://app.shortcut.com/validmind/story/5869/developer-should-be-able-to-log-unit-metric-results-to-the-backend

Associated backend PR here: (https://github.com/validmind/backend/pull/774).

This PR adds the first pass of support for unit metrics as first-class citizens. This means that unit metrics can now be run and logged just as normal tests. These unit metrics are logged to a different endpoint and saved to a different table since they are designed to be used for results over time and other platform use-cases. Each unit metric is required to be a scalar value (int or float) and is logged with a `recorded_at` timestamp to support time-based querying.

### Changes

Three high-level changes have been made by this PR:

1) The API client has been updated with a new method for logging unit metrics to the associated backend PR.
2) The result wrapper class has been updated to support scalar "unit" metrics and use the new method for logging unit metrics.
3) The Unit Metrics notebook has been updated to reflect the above changes.

Couple of smaller changes:

1) A new `list_metrics()` function has been added (and demonstrated in the notebook) for listing unit metrics
2) Current unit metrics have had the `sklearn` part removed from their path since they will work with any type of model and not just sklearn (happy to roll this back if we feel it doesn't provide more clarity)

See the below AI summary for a detailed explanation of the changes in this PR.


cachafla:
> Current unit metrics have had the sklearn part removed from their path since they will work with any type of model and not just sklearn (happy to roll this back if we feel it doesn't provide more clarity)

This was done to reflect the origin of the metric but on a second thought it could make users think that it is for sklearn models. Removing from the path should be good and we can document where the metric comes from in its description.

validmind/api_client.py
```diff
@@ -555,6 +555,66 @@ def log_input(name: str, type: str, metadata: Dict[str, Any]) -> Dict[str, Any]:
         raise e
 
 
+async def alog_metric(
+    key: str,
+    value: float,
+    inputs: Optional[List[str]] = None,
+    params: Optional[Dict[str, Any]] = None,
+    recorded_at: Optional[str] = None,
```
cachafla, line 563: Where/who do you suggest sets this value?

cachafla (approved):
Awesome 👏 I left a question about `recorded_at`. I'm assuming we want to make it explicitly different to `created_at`.

juanmleng:
Awesome stuff! Just a couple of comments: 
- It might be helpful to add a small explanation in the notebook for users to understand the difference between `run_metric` and `run_test`. For instance, is there any difference between `run_metric` and `run_test` when including a single metric?
- Additionally, what happens under the hood? Does `run_test` log a metric result in the same unit metric table as an additional metric run, which can later be used for monitoring, or does it get stored somewhere else?

juanmleng (approved):
Game changer!

validmind/api_client.py
johnwalz97: I added this here since its an optional parameter that the backend accepts and is a distinct field from created_at. Reason being is I figured we might have cases where monitoring data is computed live and then batch-shipped off to the platform. For now I was leaving it here so we are aware that that option exists but happy to remove it if we feel it's not necessary.

johnwalz97:
> Awesome stuff! Just a couple of comments: 
> 
> - It might be helpful to add a small explanation in the notebook for users to understand the difference between `run_metric` and `run_test`. For instance, is there any difference between `run_metric` and `run_test` when including a single metric?
> 
> - Additionally, what happens under the hood? Does `run_test` log a metric result in the same unit metric table as an additional metric run, which can later be used for monitoring, or does it get stored somewhere else?

- Yeah great questions @juanmleng... Right now they are essentially the same - calling run_test just calls run_metric and you will get the same return object back. Either  one will work but I made it separate in the notebook since right now they can't be used like normal tests and don't get saved to the template.

I was hoping to meet up with you @AnilSorathiya and @cachafla to see if we want to keep those separate so users know that there is a clear distinction between the two. Theres also the question of where we will allow users to actually use these unit metrics (like as part of a monitoring result or in all types of documentation). And if thats the case do they get added to the template so they get run automatically etc. Finally, I also made it so that any test can return a unit metric along with figures and tables. So the question there is do we want to expose that functionality or is that a bit confusing and we keep them separate... I can setup a call to discuss the above :)

johnwalz97:
Updated the notebook as per our discussion and added support for the input grid for normal unit metrics and composite metrics. Using the input_grid is only possible via the `run_test`. Also building a composite metric is only done with `run_test()`. This makes sense to me since in both of those cases you end up with a test result instead of a unit metric.

PR #178, opened 2024-08-26
Refactor add_test_description script

cachafla:

- Fixes to `add_test_description.py` so it generates descriptions more consistently.
- Updated the descriptions for the tests under `ongoing_monitoring`
- @validbeck feel free to make updates to the generated descriptions as needed since GPT is never fully accurate

Example proper format of a description:

---
Evaluates changes in feature distribution over time to identify potential model drift.

**Purpose**:
- To evaluate how much the distribution of features has shifted over time between two datasets, typically training
and monitoring datasets.
- To provide insights into the model's robustness and the necessity for retraining or feature engineering.

**Test Mechanism**:
- Bucketing the distributions of each feature in both datasets.
- Comparing the percentage of observations in each bucket between the two datasets.
- Aggregating the differences across all buckets for each feature to produce the Population Stability Index (PSI)
score for that feature.

**Signs of High Risk**:
- PSI >= 0.2 for any feature, indicating a significant distribution shift.
- Consistently high PSI scores across multiple features.
- Sudden spikes in PSI in recent monitoring data compared to historical data.

**Strengths**:
- Provides a quantitative measure of feature distribution changes.
- Easily interpretable thresholds for decision-making.
- Helps in early detection of data drift, prompting timely interventions.

**Limitations**:
- May not capture more intricate changes in data distribution nuances.
- Assumes that bucket thresholds (quantiles) adequately represent distribution shifts.
- PSI score interpretation can be overly simplistic for complex datasets.
---

### How to use

Entire directory:

```
poetry run python scripts/add_test_description.py review validmind/tests/ongoing_monitoring/
```

Single file:

```
poetry run python scripts/add_test_description.py review validmind/tests/ongoing_monitoring/FeatureDrift.py
```


validbeck (changes requested):
That's so cool! I tested it on files that didn't have the summary and it worked. My only request is can we maybe adjust the formatting of the output?

Can we do a `### Header 3` instead of a `**Heading**:`? 

I can go back and review the older files for this as I sort of already did them in my other branch anyway 😅 

| Old Format | New Format |
|---|---|
| Screenshot 2024-08-26 at 10 46 31 AM | Screenshot 2024-08-26 at 10 46 25 AM|

I tried to hack the script but wasn't able to enforce the new layout 🤔

validbeck (approved):
I did let's hope I didn't break everything

**EDIT:** nooo wait i broke the paragraph summaries rip hang on

PR #181, opened 2024-08-28
Edited model validation embeddings `StabilityAnalysis.py`

validbeck:

> For sc-6216 I made a copy of `StabilityAnalysis.py`, ran the `add_test_descriptions.py` on it, then manually pasted the relevant Python docstring output into the old file. 🤞🏻 HOPEFULLY the checks will not fail this time as only the docstring will have changed.

| Old | New |
|---|---|
| Screenshot 2024-08-28 at 8 42 34 AM | Screenshot 2024-08-28 at 9 01 35 AM |


validbeck:
CHECK PASSSSED 🎉

cachafla (approved):
Thanks for the fix 👍. Note that we don't want to document/expose this test description on the docs site because this test cannot be executed on its own. It's a subclass imported by other tests.

validbeck:
> Thanks for the fix 👍. Note that we don't want to document/expose this test description on the docs site because this test cannot be executed on its own. It's a subclass imported by other tests.

Interesting — I am not sure how to remove it as it gets pulled in automatically. 🤔 See here: https://docs.validmind.ai/tests/model_validation/embeddings/StabilityAnalysis.html

Maybe we need a list of tests that SHOULDN'T be retrieved by [our `make get-source`](https://github.com/validmind/documentation/blob/1dd8c468530df7b050fbfa7f7c22c953e2136f3b/site/Makefile#L17) and we can go from there?

PR #185, opened 2024-09-11
John6797/sc 6242/store existing prompts in a central location

johnwalz97:

Moving the test description generation structure from the ai risk engineering repo over to here. This is just the start and eventually we need to get the tasks library deployed to pypi so we can use it in the dev framework and backend


cachafla (approved):
Excellent👌

PR #190, opened 2024-09-18
Fix Broken RAGAS Test and Add New

johnwalz97:

@MichaelIngvarRoenning  ran into an issue where after getting a fresh install of validmind, the aspect critique test stopped working for him. This was due to RAGAS changing how that test worked. This PR fixes that by updating the minimum `ragas` version and fixing the test logic. It also adds two new tests to add some of the new tests that they have added, namely:
* Context Utilization (a reference-free version of Context Precision): https://docs.ragas.io/en/latest/concepts/metrics/context_utilization.html
* NoiseSensitivity (a new metric that tries to determine sensitivity of an LLM to noise in the context data): https://docs.ragas.io/en/latest/concepts/metrics/noise_sensitivity.html


validmind/tests/model_validation/ragas/ContextUtilization.py
```diff
@@ -0,0 +1,125 @@
...
+from datasets import Dataset
+
+from validmind import tags, tasks
+
+from .utils import get_ragas_config, get_renamed_columns
+
+
+@tags("ragas", "llm", "retrieval_performance")
+@tasks("text_qa", "text_generation", "text_summarization", "text_classification")
+def ContextUtilization(
```
cachafla, line 17: Can you pass this description through the test description script so its format is consistent with the others?

validmind/tests/model_validation/ragas/NoiseSensitivity.py
```diff
@@ -0,0 +1,132 @@
...
+
+@tags("ragas", "llm", "rag_performance")
+@tasks("text_qa", "text_generation", "text_summarization")
+def NoiseSensitivity(
+    dataset,
+    answer_column="answer",
+    contexts_column="contexts",
+    ground_truth_column="ground_truth",
+):
+    """
```
cachafla, line 23: Same here

cachafla (approved):
Awesome 👌 I left a comment on updating the test descriptions.

validmind/tests/model_validation/ragas/ContextUtilization.py
johnwalz97: yep will do

johnwalz97:
So I ran the script and it removed the instructions on running the test (i.e. the important bits about how to configure columns). Can we add a new section for that @cachafla

PR #193, opened 2024-09-20
[SC-6552] Fix tests with no outputs

juanmleng:

- Fix the following test that did not output results: 
  - `BoxPierce`
  - `LJungBox`
  - `JarqueBera`
  - `ShapiroWilk`
  - `RunsTest`
  - `DurbinWatsonTest`

- Re-write `DurbinWatsonTest` to be computed against residuals. 
- Move tests that require only `dataset` to `validmind.data_validation`.


cachafla (approved):
Nice 👌

PR #198, opened 2024-09-27
Add parameter grid support in the comparison tests functionality

AnilSorathiya:

We support for comparison tests using the `input_grid` parameter in `run_test` functionality. However,  the similar support is required for params. This is useful for any sort of case where you might want to run the same test against multiple combinations of test parameters and create a single documentation block that compares the individual results.

The updated `run_test()` function now allows you to pass an `param_grid` that will run a test for all combinations of parameters.

Example - for this param grid:

```
param_grid = {
    "param1": [1],
    "param2": [0.1, 0.2],
}
```

A test will run once for each of the following param groups:

```
{"param1": 1, "param2": 0.1}
{"param1": 1, "param2": 0.2}
```

## External Release Notes

We support for comparison tests using the `input_grid` parameter in `run_test` functionality. However,  the similar support is required for params. This is useful for any sort of case where you might want to run the same test against multiple combinations of test parameters and create a single documentation block that compares the individual results.

The updated `run_test()` function now allows you to pass an `param_grid` that will run a test for all combinations of parameters.

Example - for this param grid:

```
param_grid = {
    "param1": [1],
    "param2": [0.1, 0.2],
}
```

A test will run once for each of the following param groups:

```
{"param1": 1, "param2": 0.1}
{"param1": 1, "param2": 0.2}
```

### DIFF
--- a/notebooks/code_samples/regression/quickstart_regression_full_suite.ipynb
+++ b/notebooks/code_samples/regression/quickstart_regression_full_suite.ipynb
@@ -59,23 +59,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "id": "6G5-kHOZ7YWk"
    },
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "\u001b[33mWARNING: visions 0.7.5 does not provide the extra 'type-image-path'\u001b[0m\u001b[33m\n",
-      "\u001b[0m\n",
-      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m A new release of pip is available: \u001b[0m\u001b[31;49m23.3.1\u001b[0m\u001b[39;49m -> \u001b[0m\u001b[32;49m24.2\u001b[0m\n",
-      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m To update, run: \u001b[0m\u001b[32;49mpip install --upgrade pip\u001b[0m\n",
-      "Note: you may need to restart the kernel to use updated packages.\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "%pip install -q validmind"
    ]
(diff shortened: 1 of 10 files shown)


validmind/tests/run.py
```diff
@@ -385,83 +403,42 @@ def run_test(
     __generate_description: bool = True,
     **kwargs,
 ) -> Union[MetricResultWrapper, ThresholdTestResultWrapper]:
-    """Run a test by test ID
+    """Run a test by test ID."""
 
-    Args:
```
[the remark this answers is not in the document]
AnilSorathiya: good catch! Thanks

notebooks/code_samples/regression/quickstart_regression_full_suite.ipynb
cachafla, line 1: Seems like the changes to this notebooks are not needed?

```
import logging
import sys
import site

logger = logging.getLogger()

# Print the path to the current Python interpreter
logger.info("Python executable path: " + sys.executable)

# Print the path to the site-packages directory
logger.info("Site-packages path: " + str(site.getsitepackages()))
```

Also, the snippet was updated to use `project` instead of `model`.

notebooks/how_to/run_tests/2_run_comparison_tests.ipynb
```diff
@@ -40,7 +40,8 @@
     "  - [Run the test with one model](#toc5_5_)      \n",
     "- [Run comparison tests](#toc6_)    \n",
     "  - [Run a comparison test with multiple models](#toc6_1_)    \n",
-    "  - [Run a classifier performance comparison test](#toc6_2_)      \n",
+    "  - [Run a comparison test with multiple parameters' value](#toc6_2_)    \n",
```
cachafla, line 43:
```suggestion
    "  - [Run a comparison test with multiple parameters' values](#toc6_2_)    \n",
```

validmind/tests/run.py
```diff
@@ -385,83 +403,81 @@ def run_test(
...
-            - dataset: A validmind Dataset object or a Pandas DataFrame
-            - model: A model to use for the test
-            - models: A list of models to use for the test
-            - dataset: A validmind Dataset object or a Pandas DataFrame
+    """Run a test by test ID.
+    test_id (TestID, optional): The test ID to run. Not required if `unit_metrics` is provided.
+    params (dict, optional): A dictionary of parameters to pass into the test. Params
+        are used to customize the test behavior and are specific to each test. See the
+        test details for more information on the available parameters. Defaults to None.
+    input_grid (Union[Dict[str, List[Any]], List[Dict[str, Any]]], optional): To run
```
cachafla, line 411:
```suggestion
    param_grid (Union[Dict[str, List[Any]], List[Dict[str, Any]]], optional): To run
```

cachafla (approved):
Nice! 👏

cachafla:
Left two comments to ensure 1) we don't commit an unrelated notebook and 2) the docstring is updated with `param_grid` instead of `input_grid`.

validmind/tests/run.py
AnilSorathiya: good catch!

PR #200, opened 2024-10-04
Fix PyPI publish

cachafla:

It started to fail randomly with a missing license error so we might as well start using `poetry publish` and ensure we use the latest version of twine.


cachafla:
This was already tested and it worked here: https://pypi.org/project/validmind/2.5.18/.

juanmleng (approved):

PR #202, opened 2024-10-04
John6797/sc 6723/step 1 get rid of test runs

johnwalz97:

Removing test runs.


cachafla (approved):
Tested, works great.

PR #205, opened 2024-10-09
Require aequitas at runtime

cachafla:

Not a necessary fix but this is an annoying condition where `aequitas` is required even to run `list_tests()`. @johnwalz97 the current issue with this library is that the `lightgbm` doesn't install (compile) cleanly on Mac and there are multiple hacks required in order to make it work. This is also why `aequitas` is not on `pyproject.toml` since it won't compile cleanly.

While we figure out a good solution to this I propose we only load the library at runtime so it will still work on JupyterHub et al.

### DIFF
--- a/validmind/tests/data_validation/ProtectedClassesDisparity.py
+++ b/validmind/tests/data_validation/ProtectedClassesDisparity.py
@@ -5,11 +5,7 @@
 import io
 import sys
 
-import aequitas.plot as ap
 import pandas as pd
-from aequitas.bias import Bias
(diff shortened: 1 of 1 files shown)


johnwalz97 (approved):
thanks... have a more formal fix for this kind of situation locally that I can push up soon but this is a good short-term solution

juanmleng (approved):

cachafla:
Superseded by https://github.com/validmind/developer-framework/pull/206

PR #206, opened 2024-10-15
fix: fixing issue with missing dependency

johnwalz97:

Adding logic to handle tests that have missing dependencies in a standardized way. This will let users list and load tests without having the dependencies those tests require but it will notify them of this so they aren't missing out on any tests.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,12 +13,13 @@ readme = "README.pypi.md"
 version = "2.5.19"
 
 [tool.poetry.dependencies]
+python = ">=3.8.1,<3.12"
 aiohttp = {extras = ["speedups"], version = "*"}
 arch = "*"
 bert-score = ">=0.3.13"
 catboost = "*"
-evaluate = "*"
 datasets = "^2.10.0"
+evaluate = "*"
 ipywidgets = "*"
 kaleido = ">=0.2.1,!=0.2.1.post1"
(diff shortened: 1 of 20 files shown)


cachafla:
Oops I forgot to merge this other PR 😬 https://github.com/validmind/developer-framework/pull/205. I won't do it anymore since you added this @johnwalz97

pyproject.toml
```diff
@@ -13,12 +13,13 @@ readme = "README.pypi.md"
 version = "2.5.19"
 
 [tool.poetry.dependencies]
+python = ">=3.8.1,<3.12"
```
cachafla, line 16: This will keep moving on some of us' editor formatters that have auto-sort dependencies enabled 😅

cachafla (approved):
Awesome 👏. Could you target this PR to merge into `prod` instead?

pyproject.toml
johnwalz97: lol I've been meaning to add something to the readme to install the poetry auto-sort plugin so we all have the same auto-sorter 😂

johnwalz97:
> Oops I forgot to merge this other PR 😬 #205. I won't do it anymore since you added this @johnwalz97

all good... either technically works to fix the issue but i figured this would be a more standardized approach that we can start adopting 😅

PR #212, opened 2024-10-24
Added Upgrade ValidMind instructions to Jupyter Notebooks

validbeck:

> sc-7192

### Upgrade ValidMind

Where it made sense in the notebooks, I added the upgrade information as a footer and made sure all the ToC & anchors were adjusted to add a link to this new section:

Screenshot 2024-10-24 at 11 53 22 AM

### Updated callouts

Updated the look & feel of these to our new docs site branding:

| Old | New |
|---|---|
| Screenshot 2024-10-24 at 11 55 24 AM | Screenshot 2024-10-24 at 11 53 15 AM|


cachafla (commented):
Much needed thanks 🙏.

One small change I'd recommend is to not make the show/upgrade commands code cells but "code blocks" inside Markdown (text blocks). That way we don't force users to upgrade `validmind` every time they run notebooks end to end, which is something we do a lot during 1) customer demos and 2) automated tests.

There are valid reasons for not always upgrading to the latest from upstream. Example: when running on JupyterHub we might want to use a locally modified `validmind` that has a patch not available upstream yet. If the user were to hit on "run all" they would override their local `validmind` inadvertently.

Showing the upgrade instructions as a code block in Markdown is an extra step but a small one IMO.

validbeck:
@cachafla Now they all look like this:

Screenshot 2024-10-30 at 9 47 04 AM

cachafla (approved):
Yep, not sure. I ran `make lint` locally and it works.

PR #213, opened 2024-10-28
chore: undo some testing changes

eggshell:
We can go back to ubuntu-latest and turn failure notifications back on, these were only changed during testing the integration test failures last week.


johnwalz97 (approved):
nice 👌

PR #216, opened 2024-10-31
John6797/sc 6728/table consolidation step 6 add support for

johnwalz97:

Refactoring to move all tests to the functional format and removing anything related to class-based tests or threshold tests. Also removing different ways that tests get loaded, initialized and run so we have a single code-path (the `run_test()` interface) for all types of tests and unit metrics.


juanmleng:
I noticed that some tests, such as the prompt evaluation tests, lack type annotations. This is not critical, so as we work on other PRs, we can gradually add type annotations to these tests.

AnilSorathiya (approved):
go for it! All looks good to me.

juanmleng (approved):
All looking great John. Thanks for this massive effort!

PR #229, opened 2024-11-08
fix: run integration tests on large runners

eggshell:
This is something I'd been going back and forth on, but now we've got a reason to use it. With all the large dependencies in the developer framework, we are eclipsing the relatively small amount of storage space on GitHub's standard runners. We already have a pool of larger runners being used to build our docker images faster, but they also have more disk space. If we notice jobs being queued, we can always add more large runners to the pool

## External Release Notes


juanmleng (approved):
Awesome. Thanks for fixing this!

cachafla (approved):

PR #232, opened 2024-11-13
[SC 7399] Serialize date, time, datetime and Date objects to log test results 

AnilSorathiya:
Error while logging the date, time, datetime or QuantLib.Date object in the test result:
```python
File ~/work/validmind/repos/developer-framework/validmind/utils.py:111, in NumpyEncoder.default(self, obj)
    109 if isinstance(obj, set):
    110     return list(obj)
--> 111 return super().default(obj)

File /opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/json/encoder.py:179, in JSONEncoder.default(self, o)
    160 def default(self, o):
    161     """Implement this method in a subclass such that it returns
    162     a serializable object for ``o``, or calls the base implementation
    163     (to raise a ``TypeError``).
   (...)
    177 
    178     """
--> 179     raise TypeError(f'Object of type {o.__class__.__name__} '
    180                     f'is not JSON serializable')

TypeError: Object of type Date is not JSON serializable
```

### Solution:
Convert into string format to serialize date/time type object.


validmind/utils.py
```diff
@@ -106,6 +110,9 @@ def default(self, obj):
             return bool(obj)
         if isinstance(obj, pd.Timestamp):
             return str(obj)
+        if isinstance(obj, ql.Date):
+            # Convert QuantLib Date to string in YYYY-MM-DD format
+            return f"{obj.year()}-{obj.month()}-{obj.dayOfMonth()}"
```
cachafla, line 115: You can use ISO() here:

```py
>>> start = ql.Date(7,5,2020)
>>> start.ISO()
'2020-05-07'
```

AnilSorathiya: thanks

cachafla (approved):
Nice 👍

johnwalz97 (approved):
nice 👍

PR #244, opened 2024-11-25
[SC 7429] Add title as an optional parameter in the run_test

AnilSorathiya:


validmind/vm_models/result/result.py
```diff
@@ -179,13 +185,14 @@ def to_widget(self):
             self._was_description_generated = True
 
         if self.metric is not None and not self.tables and not self.figures:
-            return HTML(f"<h3>{self.result_id}: <code>{self.metric}</code></h3>")
+            return HTML(
+                f"<h3>{self.title or test_id_to_name(self.result_id)}: <code>{self.metric}</code></h3>"
```
cachafla, line 189: Given the above new property this should be:


```suggestion
                f"<h3>{self.test_name}: <code>{self.metric}</code></h3>"
```

validmind/vm_models/result/result.py
```diff
@@ -179,13 +185,14 @@ def to_widget(self):
...
+            )
 
         template_data = {
-            "test_name": test_id_to_name(self.result_id),
+            "test_name": self.title or test_id_to_name(self.result_id),
```
cachafla, line 193:
```suggestion
            "test_name": self.test_name,
```

cachafla (approved):
Looks good 👌. I left a change suggestion to make use of the new property that was added.

PR #248, opened 2024-11-26
[SC-7553] Fix error in comparison tests when not using parameters

juanmleng:
The `combine_results()` function in `validmind/tests/comparison.py` fails with an `AttributeError` when processing test results that don't have parameters. 

Fix: Add null checks before accessing `result.params` to handle running comparison tests that don't specify parameters.

I also moved the import of XGBoost to the top in a separate cell in `2_run_comparison_tests.ipynb`, as the model fitting was causing a strange Jupyter crash. However, when I moved it to a separate cell, it worked. Not sure why, though.


johnwalz97 (approved):
LGTM... Just ran into this as well. Thanks for the fix!

PR #252, opened 2024-11-27
chore: removing metrics references

johnwalz97:

Removing any references to old metrics.


cachafla (approved):
That's it? 😅

PR #253, opened 2024-11-28
[SC-7605] Fix: input data mutation in PSI test affecting other tests during documentation suite execution

juanmleng:

A bug was identified in the documentation suite in the customer churn notebook, where the PSI test was modifying the input data of the ROC test. This caused the ROC test to output an unexpected value and curve. The issue originated in the `calculate_psi` function, which directly sorted the input data (`score_initial` and `score_new`) without creating independent copies.

The fix is ensuring the PSI function works with copies of the input data instead of the original. After implementing this fix, the suite executed correctly.


cachafla (approved):
Great catch 👏

PR #256, opened 2024-12-02
[SC-7588] Explore dynamic llm context injection in test descriptions

juanmleng:

We aim to add test-specific context into the LLM-based result descriptions for individual tests to improve the quality of the insights and align the descriptions with use-case and bank policy requirements. This PR starts this effort by experimenting with setting global context using an environment variable in the notebook, which will be used to enhance the user prompt with external context related to the use case.


johnwalz97 (approved):
Very nice! Looking forward to playing around with this!

cachafla (approved):
👏

PR #263, opened 2024-12-10
Add Initial Support for Post-Processing Functions

johnwalz97:

Initial support for post-processing functions along with an example notebook with a lot of examples.

## External Release Notes

Adding support for post-processing functions to allow users to modify test outputs without modifying test code.


AnilSorathiya (commented):
@johnwalz97 Awesome 👍  
There are a couple of things:
1. Are we using `key` anywhere? If not why don't we create separate ticket get rid off it?
2. Do we need `ref_id` for `RawData` to access it on platform/frontend? Since we are using `ref_id` for tables and figures.

johnwalz97:
> @johnwalz97 Awesome 👍  
> 
> There are a couple of things:
> 
> 1. Are we using `key` anywhere? If not why don't we create separate ticket get rid off it?
> 
> 2. Do we need `ref_id` for `RawData` to access it on platform/frontend? Since we are using `ref_id` for tables and figures.

Thanks for reviewing @AnilSorathiya 

1. Thats a great idea, I can add a ticket for that.
2. We don't need it yet since ref_id is only used to link figures to test results on the backend (since they are logged stored separately in the DB). If we ever start storing raw_data outside of the test_results table than we can add it.

notebooks/code_sharing/post_processing_functions.ipynb
AnilSorathiya, line 1: just for reference, 
Can you add an example of accessing the RawData from a test result and create plot/table out of it?

AnilSorathiya (approved):
left one last comment otherwise it looks good.

PR #265, opened 2024-12-11
Hotfix/fix listing of tasks and tags

johnwalz97:


cachafla (approved):
If this is something the "explore tests" notebook would have made evident then let's add that notebook to the integration tests GH action.

johnwalz97:
> If this is something the "explore tests" notebook would have made evident then let's add that notebook to the integration tests GH action.

 It would have @cachafla... Adding that in my PR to main.

PR #267, opened 2024-12-11
Bump Version to 2.6.9

johnwalz97:

Apologies @cachafla and @AnilSorathiya for yet another hotfix PR. I didn't bump the version on the last one so the release failed due to version conflicts when pushing to PyPi. I was hoping that it would overwrite the old one but a special flag has to be set for that. Bumping is the better practice in any case so this should fix that.


validbeck (approved):
Thank you SO much for hopping on this!!!

PR #272, opened 2024-12-12
[SC 7726] Create graphs in the capital markets model

AnilSorathiya:
* Create graphs from test output using post_process function


johnwalz97 (approved):
great stuff... let me merge mine first and then you should be able to merge main to sync up with those changes

PR #274, opened 2024-12-13
[SC-7759] Exposing static descriptions in test results

juanmleng:

Added a `doc` property to `TestResult` class to clearly separate the static test documentation from the LLM-generated descriptions. This change helps distinguish between the original test documentation (accessed via `result.doc`) and the dynamic, descriptions generated by the LLM (accessed via `result.description`). The `doc` property contains the test's docstring, providing users with direct access to the test's purpose and behavior.


johnwalz97 (approved):
lgtm 👏

PR #276, opened 2024-12-16
feat: test [all] install target in integration tests

nibalizer:


nibalizer:
This doesn't make sense because we don't have the R dependencies in this test.

PR #285, opened 2025-01-07
John6797/sc 7792/add raw data to validmind library tests

johnwalz97:


tests/test_integration_tests.py
juanmleng, line 1: Is the idea here a permanent or temporary solution to remove this testing and run only unit tests?

juanmleng (approved):
Really cool and extremely useful job @johnwalz97! Just left a minor comment. Otherwise looks good to go from my side.

tests/test_integration_tests.py
cachafla: Is this being removed on purpose? 😬

johnwalz97: Yep, I removed that intentionally. Reasoning there was since we now have @juanmleng's unit tests, theres no reason to run tests twice. AFAIK, this script merely runs the tests and checks that they don't throw an error whereas the new unit tests actually test for correct outputs.

AnilSorathiya (approved):
Thanks John. LGTM!

tests/test_integration_tests.py
cachafla: Sounds good. Then let's remove this file completely :).

cachafla (approved):

tests/test_integration_tests.py
johnwalz97: sounds good 👍

PR #294, opened 2025-01-14
John6797/sc 8121/switch from mathml to mathjax for latex formulas

johnwalz97:

Refactoring to use mathjax instead of mathml in test description html. This aligns with the new math plugin we are using on the frontend.


cachafla (approved):
Nice 👌

PR #296, opened 2025-01-17
John6797/sc 7670/move test result description prompts to backend

johnwalz97:

Moving test result description generation to the backend.


juanmleng:
Would this change still support passing external context via `VALIDMIND_LLM_DESCRIPTIONS_CONTEXT`?

pyproject.toml
```diff
@@ -10,7 +10,7 @@ description = "ValidMind Library"
 license = "Commercial License"
 name = "validmind"
 readme = "README.pypi.md"
-version = "2.7.7"
+version = "2.7.8"
```
cachafla, line 13: Let's go with 2.8.0 since this is a meaningful change 🙂

cachafla:
> Would this change still support passing external context via `VALIDMIND_LLM_DESCRIPTIONS_CONTEXT`?

The `context` is still being passed to the request, except that now the request is sent to the API and unpacked here: https://github.com/validmind/backend/pull/1140/files#diff-1f286300bd4ef710c7d4c921c3702e65de730379fc39f4b68c686cc4d92c1345R412

If you check the the `generate_description()` function you'll see that `context` is still being retrieved in the same way:

```
context = _get_llm_global_context()
```

cachafla (approved):
Awesome 👍

validmind/ai/test_descriptions.py
```diff
@@ -131,19 +131,22 @@ def generate_description(
...
+    # system, user = _load_prompt()
+
+    # messages = [
+    #     prompt_to_message("system", system.render(input_data)),
+    #     prompt_to_message("user", user.render(input_data)),
+    # ]
+    # response = client.chat.completions.create(
+    #     model=model,
+    #     temperature=0.0,
+    #     messages=messages,
```
AnilSorathiya, line 143: why do we have commented code here?

AnilSorathiya (approved):
left one comment. Otherwise it is LGTM 👍

pyproject.toml
johnwalz97: sounds good

validmind/ai/test_descriptions.py
johnwalz97: in case we want to still enable basic test result descriptions in the library for those who aren't connected to the API or who want to use their own LLM directly

PR #300, opened 2025-01-27
John6797/sc 8257/regularly check if the user is using an older

johnwalz97:

Add background check to see if kernel is running the latest installed validmind version.


cachafla (approved):
Awesome 👏

PR #302, opened 2025-01-29
[SC-8300] Fix: Store embeddings in dictionary instead of DataFrame

juanmleng:
Some embedding tests were failing because we were trying to store multi-dimensional embedding arrays in the `RawData` using a DataFrame. The fix changes the raw data storage from DataFrame to dictionary to properly store the embeddings.


johnwalz97 (approved):
thanks for fixing this Juan!

PR #304, opened 2025-01-29
Fix: Adding Missing `RawData` Support to Comparison Tests

johnwalz97:

The newly added raw data feature for tests is not currently supported when running comparison tests (i.e. using an input_grid or param_grid). For this initial solution, I am just appending the raw data elements to a list.


validmind/tests/comparison.py
```diff
@@ -312,6 +312,25 @@ def list_to_dict(grid_list):
     return test_configs
 
 
+def _combine_raw_data(results: List[TestResult]) -> RawData:
+    """Combine RawData objects"""
```
juanmleng, line 316: To avoid failing tests when raw data is missing, we can return an empty RawData object in those cases:

```
def _combine_raw_data(results: List[TestResult]) -> RawData:
    """Combine RawData objects"""
    # Handle case where result has no raw data
    if not results[0].raw_data:
        return RawData()
```

juanmleng (approved):
Looks great, John! Thanks for addressing this. I just left a suggestion to handle cases where the tests don´t return raw data.

validmind/tests/comparison.py
johnwalz97: good callout @juanmleng... fixed that :)

PR #305, opened 2025-01-30
Forgot an end div omg

validbeck:

Forgot an end div in the "Log metrics over time" notebook causing display issues when rendered in docs 🤦🏻


nrichers (approved):
LGTM! 🤖

PR #307, opened 2025-01-31
Edited adding context to LLM-generated descriptions notebook

validbeck:

> sc-8278

Part 1 for getting ready to expose information about adding context to LLM-generated test descriptions. I updated the `llm_descriptions_context.ipynb` (renamed: `add_context_to_llm_descriptions.ipynb`) with the idea that it will eventually be moved into "how_to": https://github.com/validmind/validmind-library/blob/beck/sc-8278/documentation-spike-explore-dynamic-llm-context/notebooks/code_sharing/llm/add_context_to_llm_descriptions.ipynb

I will check in with Kristof and Mehdi to see if we want to expose this notebook to customers and update some of the documentation outside of the notebook to signal this use case.


juanmleng (approved):
Looks great to me Beck! Thanks for the edits :)

PR #309, opened 2025-02-04
Notebook to show Raw Data Feature

johnwalz97:

Adding a notebook to directly demo the new Raw Data feature as well as a quick fix to better display raw data via the `inspect()` function.


validbeck (approved):
Thanks John! I cleaned up the notebook, applied our style guidelines, and moved it into `how_to`. Once this PR is merged I'll also wrap the notebook into our docs site.

PR #310, opened 2025-02-07
[SC-8445] fix: Apply post_process_fn to individual results in comparison tests

juanmleng:

When running comparison tests with `input_grid` or `param_grid` and a `post_process_fn`, the post-processing function is applied to the combined results rather than to individual test results. This can lead to an unexpected or unintuitive display of outputs, as post-processing functions are not associated with a specific `input_id` but are instead applied to the aggregated results.

### DIFF
--- a/validmind/tests/run.py
+++ b/validmind/tests/run.py
@@ -222,6 +222,7 @@ def _run_comparison_test(
     params: Union[Dict[str, Any], None],
     param_grid: Union[Dict[str, List[Any]], List[Dict[str, Any]], None],
     title: Optional[str] = None,
+    post_process_fn: Optional[Callable[[TestResult], None]] = None,
 ):
     """Run a comparison test i.e. a test that compares multiple outputs of a test across
     different input and/or param combinations"""
@@ -232,8 +233,9 @@ def _run_comparison_test(
         params=params,
     )
 
-    results = [
-        run_test(
+    results = []
+    for config in run_test_configs:
+        result = run_test(
             test_id=test_id,
             name=name,
             unit_metrics=unit_metrics,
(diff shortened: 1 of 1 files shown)


validmind/tests/run.py
johnwalz97, line 1: I am not sure that I agree with the choice to run post-process functions for each individual result in a comparison test instead of on the final result. Although it does make sense to have the ability to do so. Maybe we should discuss in vm-library channel to see what the team thinks?

johnwalz97 (changes requested):

validmind/tests/run.py
juanmleng: @johnwalz97 raised a very valid point here, and made me think about this a bit more. I totally agree @AnilSorathiya If we add the `input_id` to the `RawData` in our tests and leave the user to handle the combined and individual results in the post-processing, we can indeed get to the desired results.

juanmleng:
Closing this PR as we've decided that maintaining the current API is the best approach.

Next steps:
1. Create a new PR to add input IDs to RawData in our tests
2. This will better support post-processing in comparison tests
3. Document post-processing and RawData for comparison tests in our how-to guides

The existing `RawData` + `processing_fn` architecture already provides the flexibility we need, we just need to leverage it better.

juanmleng:
Closing this PR as we've decided that maintaining the current API is the best approach.

Next steps:
1. Create a new PR to add input IDs to RawData in our tests
2. This will better support post-processing in comparison tests
3. Document post-processing and RawData for comparison tests in our how-to guides

The existing `RawData` + `processing_fn` architecture already provides the flexibility we need, we just need to leverage it better.

PR #311, opened 2025-02-11
fix: handle context length too large

johnwalz97:

Hotfix for issue with context length exceeded errors.


cachafla (approved):
Great 👌

PR #316, opened 2025-02-18
[SC-8240] Add warning message for unregistered tests in template

juanmleng:

Added a new method `_check_result_id_exists()` that scans all sections in the documentation template to verify if a test result ID already exists anywhere in the document. This helper method:

- Checks for duplicate test results across all sections
- Logs an info message when a test result ID is not found in any section


validmind/vm_models/result/result.py
```diff
@@ -440,6 +440,9 @@ def log(self, section_id: str = None, position: int = None, unsafe: bool = False
             unsafe (bool): If True, log the result even if it contains sensitive data
                 i.e. raw data from input datasets
         """
+
+        self.check_result_id_exist()
```
cachafla, line 444: Does this mean we'll call this method once for every log when we run `run_documentation_tests()`? I'm concerned about running `api_client.reload()` N times (one per test in the template).

I wonder if there's a place where we can we reload the template only once for these two situations:
- When logging and individual result via `result.log()`
- When running `run_documentation_tests()` which in turns triggers a bunch of individual `result.log()` calls

juanmleng: Great call out. Looking at run_documentation_tests() my understanding is that follows the log_async() path, not the result.log(), so it would not trigger multiple reloads through check_result_id_exist(). However, you raised a valid concern about reloading the template multiple times. To address this, I´ve added a chaching logic, to ensure that for individual result.log() and result.log(section_id=<section_id>) calls, the client_config is loaded once and cached. Let me know if this makes sense, or if I missed something.

cachafla (approved):
Nice 👌

PR #320, opened 2025-02-20
Updating Failing Custom Test Example that Relied on External Service

johnwalz97:
The BTC price API that we were using to demonstrate tests that can pull from external systems went down and is causing our integration tests notebook to fail. This PR changes the example custom test to pull from a more reliable free API to demonstrate something similar with fake user data.


cachafla (approved):
Awesome 🫡

PR #328, opened 2025-03-06
ValidMind for model development quick formatting edit

validbeck:

Quarto doesn't like that there were no line breaks before and after these code blocks, and didn't render them properly. I fixed it in the source.

| Before | After |
|---|---|
| Screenshot 2025-03-06 at 12 26 37 PM|Screenshot 2025-03-06 at 12 30 44 PM |


nrichers (approved):
LGTM 🚀

PR #333, opened 2025-03-09
[SC-8983] Remove ydata-profiling  library dependency

AnilSorathiya:
There are not much dependencies of the `ydata-profiling` library and new version showing the message which can confuse users of the vm-library.


johnwalz97 (approved):
beautiful 😍...

my only suggestion is maybe to break the utils.py file into separate files (dataset_utils.py, model_utils.py etc) since its getting quite long

cachafla (approved):
🫡

PR #335, opened 2025-03-11
[SC-9912] Fix index issue when using input_grid

juanmleng:

Fix InvalidIndexError: Non-unique indices during DataFrame concatenation. The error occurs because pandas is trying to concatenate DataFrames where the indices are not unique, which happens when we're combining results from multiple features and their bins.


cachafla (approved):
Nice

PR #336, opened 2025-03-11
[SC 8983] Remove inferring datatype warnings

AnilSorathiya:
Remove warning message:
```
/../../developer-framework/validmind/utils.py:639: UserWarning:

This pattern is interpreted as a regular expression, and has match groups. To actually get the groups, use str.extract.
```


johnwalz97 (approved):
nice!

PR #337, opened 2025-03-13
[SC 8805] support config option for logging test results

AnilSorathiya:
We now have configuration options that can be passed through `result.log()` to display and customize the test results block in UI documents.

Available config options:
    - hideTitle
    - hideText
    - hideParams
    - hideTables
    - hideFigures
   
```python
test = vm.tests.run_test(
    "validmind.data_validation.TabularDescriptionTables:raw_dataset",
    input_grid={
        "dataset": [vm_raw_dataset],
    },
)
test.log(config={"hideFigures": False, "hideTables": True})
```

## External Release Notes
We now have configuration options that can be passed through `result.log()` to display and customize the test results block in UI documents.

Available config options:
    - hideTitle
    - hideText
    - hideParams
    - hideTables
    - hideFigures
   
```python
test = vm.tests.run_test(
    "validmind.data_validation.TabularDescriptionTables:raw_dataset",
    input_grid={
        "dataset": [vm_raw_dataset],
    },
)
test.log(config={"hideFigures": False, "hideTables": True})
```

### DIFF
--- a/validmind/api_client.py
+++ b/validmind/api_client.py
@@ -330,6 +330,8 @@ async def alog_test_result(
     result: Dict[str, Any],
     section_id: str = None,
     position: int = None,
+    unsafe: bool = False,
+    config: Dict[str, bool] = None,
 ) -> Dict[str, Any]:
     """Logs test results information
 
@@ -357,7 +359,7 @@ async def alog_test_result(
             "log_test_results",
             params=request_params,
             data=json.dumps(
-                result,
+                {**result, "config": config},
                 cls=NumpyEncoder,
                 allow_nan=False,
             ),
--- a/validmind/api_client.py
+++ b/validmind/api_client.py
@@ -412,7 +412,6 @@ async def alog_metric(
     value: Union[int, float],
     inputs: Optional[List[str]] = None,
     params: Optional[Dict[str, Any]] = None,
-    config: Optional[Dict[str, Any]] = None,
     recorded_at: Optional[str] = None,
     thresholds: Optional[Dict[str, Any]] = None,
 ):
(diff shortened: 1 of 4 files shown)


validmind/api_client.py
```diff
@@ -455,27 +459,22 @@ def log_metric(
     value: float,
     inputs: Optional[List[str]] = None,
     params: Optional[Dict[str, Any]] = None,
+    config: Optional[Dict[str, Any]] = None,
     recorded_at: Optional[str] = None,
     thresholds: Optional[Dict[str, Any]] = None,
 ):
-    """Logs a unit metric
```
johnwalz97, line 461: Left a comment on the backend PR but just wanted to confirm that this actually works. I had thought that the metric is a separate entitity than a metric over time block so I am not sure if this will actually apply.

johnwalz97 (approved):
Just a couple of questions but other than that lgtm

validmind/api_client.py
```diff
@@ -330,6 +330,8 @@ async def alog_test_result(
     result: Dict[str, Any],
     section_id: str = None,
     position: int = None,
+    unsafe: bool = False,
```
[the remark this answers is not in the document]
AnilSorathiya: good catch, this is not used in the backend. thanks

validmind/api_client.py
AnilSorathiya: You are right. It will apply if we introduce test description in the metric over time block.

validmind/api_client.py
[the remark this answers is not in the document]
AnilSorathiya: removed `unsafe` parameter where we don't needed.

cachafla (approved):
Nice, looks good 👍

validmind/api_client.py
AnilSorathiya: As it was suggested in the backend PR. I have removed this. We will add later when we need it.

johnwalz97 (approved):
Nice!

PR #345, opened 2025-04-02
[SC-9442] Fix param_grid showing duplicated combinations and missing parameters in figure titles

juanmleng:
**Fix for `param_grid` list of dictionaries**
Fixed a bug where passing a list of dictionaries to `param_grid` would incorrectly generate a cartesian product of all parameter values. Now, each dictionary in the list is correctly treated as a distinct test configuration, eliminating duplicated test combinations.

**Allow `param_grid` values to be displayed in figure titles**
Added a new `show_params` option to `run_test()` that enables parameter values to be displayed in figure titles for comparison tests. When enabled, figure titles now include parameter information, making it easier to distinguish between different test configurations in visualizations. The default value is `show_params=True`. If parameter values are causing the titles to become cluttered or difficult to read, the user can set `show_params=False` to prevent these values from being displayed.

### DIFF
--- a/validmind/tests/comparison.py
+++ b/validmind/tests/comparison.py
@@ -146,7 +146,9 @@ def _combine_tables(results: List[TestResult]) -> List[pd.DataFrame]:
     return [_combine_single_table(results, i) for i in range(len(results[0].tables))]
 
 
-def _build_input_param_string(result: TestResult, results: List[TestResult]) -> str:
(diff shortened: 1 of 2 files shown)


cachafla (approved):
Looks good 🫡

johnwalz97 (approved):
nice 👍

PR #349, opened 2025-04-09
[SC 9567] Add text support in the test result object

AnilSorathiya:
## External Release Notes

We now support text output in test results.
If a test returns a string, this string will be used as the test description. In such cases, the automatic generation of the test result description will be skipped, and the test output will be used directly.

```python
@vm.test("my_custom_tests.MyCustomTest")
def my_custom_test(dataset, model):
    """
    This is a custom test that does nothing.
    """
    y_true = dataset.y
    y_pred = dataset.y_pred(model)

    confusion_matrix = metrics.confusion_matrix(y_true, y_pred)

    cm_display = metrics.ConfusionMatrixDisplay(
        confusion_matrix=confusion_matrix, display_labels=[False, True]
    )
    cm_display.plot()

    plt.close()  # close the plot to avoid displaying it



    return cm_display.figure_, "Test Description - Confusion Matrix", pd.DataFrame({"Value": [1, 2, 3]})  # return the figure object itself
```

### Run test

```python

result = run_test(
    "my_custom_tests.MyCustomTest",
    inputs={"model": "model", "dataset": "test_dataset"},
)
```

### Output

image

### DIFF
--- a/validmind/tests/output.py
+++ b/validmind/tests/output.py
@@ -9,6 +9,7 @@ from uuid import uuid4
 import numpy as np
 import pandas as pd
 
+from validmind.utils import is_html, md_to_html
 from validmind.vm_models.figure import (
     Figure,
     is_matplotlib_figure,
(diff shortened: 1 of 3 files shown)


juanmleng:
Looks great. Perhaps worth adding this example to `notebooks/code_samples/custom_tests/implement_custom_tests.ipynb`

juanmleng (approved):

cachafla:
Late to the party but very nice 😊

PR #355, opened 2025-04-23
[SC 9539] Keep order of the sections while previewing the template

AnilSorathiya:
When you call `preview_template()` for model documentation, validation reports, or ongoing monitoring reports, it will spit out the preview with the sections ordered as they are displayed in the YAML instead of honouring the custom `order` specified. **For example:**

| Custom Template Order | Incorrect Preview |
|---|---|
| image| image|

This change will sort the section tree by the order of the sections in the template (if provided) and set the order to 9999 for the sections that do not have an order.


johnwalz97 (approved):
nice 👌

PR #356, opened 2025-04-23
[SC-9919] Fix pandas DataFrame dtype preservation in VMDataset initialization

juanmleng:
This PR addresses an issue where some pandas `DataFrame` `dtypes` (e.g., categorical types) are lost during `VMDataset` initialization. This change bypasses the VMDataset initialization to modify only the `DataFrameDataset` class to store the original `DataFrame` directly instead of converting to `NumPy` arrays and back. This ensures that all pandas-specific `dtype` information and metadata are preserved. 

*BEFORE:*
Screenshot 2025-04-22 at 22 21 00

*AFTER:*
Screenshot 2025-04-23 at 14 03 46

**Testing**
Run successfully `quickstart_customer_churn_full_suite.ipynb` and the `application_scorecard_executive.ipynb` notebooks.


validmind/vm_models/dataset/dataset.py
```diff
@@ -537,24 +537,34 @@ def __init__(
...
-            columns=raw_dataset.columns.to_list(),
-            target_column=target_column,
-            extra_columns=extra_columns,
-            feature_columns=feature_columns,
-            text_column=text_column,
-            target_class_labels=target_class_labels,
-            date_time_index=date_time_index,
-        )
+        # Store the DataFrame directly
+        self._df = raw_dataset.copy()
```
cachafla, line 552: I think we can do the following:

```suggestion
        self._df = raw_dataset
```

Since the `.df` accessor already does this:

```python
    @property
    def df(self) -> pd.DataFrame:
        """
        Returns the dataset as a pandas DataFrame.

        Returns:
            pd.DataFrame: The dataset as a pandas DataFrame.
        """
        # only include feature, text and target columns
        # don't include internal pred and prob columns
        columns = self.feature_columns.copy()

        # text column can also be a feature column so don't add it twice
        if self.text_column and self.text_column not in columns:
            columns.append(self.text_column)

        if self.extra_columns.extras:
            # add user-defined extra columns
            columns.extend(self.extra_columns.extras)

        if self.target_column:
            # shouldn't be a feature column but add this to be safe
            assert self.target_column not in columns
            columns.append(self.target_column)

        # return a copy to prevent accidental modification
        return as_df(self._df[columns]).copy()
```

johnwalz97, line 552: Instead of making a copy, let's just set `self._df = raw_dataset`

validmind/vm_models/dataset/dataset.py
```diff
@@ -537,24 +537,34 @@ def __init__(
             target_class_labels (dict, optional): The class labels for the target columns. Defaults to None.
             date_time_index (bool, optional): Whether to use date-time index. Defaults to False.
         """
+
+        VMInput.__init__(self)
+
+        self.input_id = input_id
+        self._raw_dataset = raw_dataset.values
```
johnwalz97, line 544: I did a search of the codebase and we never actually use `_raw_dataset` except in the `__init__` of the base `VMDataset` class. I think we can safely remove this line which will prevent yet another "copy" of the dataset

johnwalz97 (dismissed):
Couple of comments to reduce memory overhead

validmind/vm_models/dataset/dataset.py
cachafla: Follow up (from our call):

As another short term fix, I suggest allowing the `init_dataset()` to have an option to not create copies under conditions of limited resources or when dealing with large datasets. The user will have to be aware that they shouldn't be making changes to these dataframes inside tests.

validmind/vm_models/dataset/dataset.py
cachafla: Great 🫡

juanmleng: Great point! Removed that extra copy.

validmind/vm_models/dataset/dataset.py
juanmleng: Great call out @johnwalz97 !

cachafla (approved):
Awesome. Let me do some testing with this notebook 🙂

cachafla:
Big memory savings:

vm_ds = vm.init_dataset(
    dataset=df,
    input_id="demo",
    target_column="target",
)

**Small dataset:**
num_rows = 500000
num_Float64 = 30
num_float64 = 30
num_int64 = 16

Before:
used 2850.8 MiB RAM in 3.40s (system mean cpu 22%, single max cpu 77%)

After:
used 478.4 MiB RAM in 1.45s (system mean cpu 18%, single max cpu 70%)

**Bigger dataset:**
num_rows = 1000000
num_Float64 = 50
num_float64 = 50
num_int64 = 30

Before:
used 7284.9 MiB RAM in 10.70s (system mean cpu 16%, single max cpu 100%)

After:
used 2469.3 MiB RAM in 3.55s (system mean cpu 15%, single max cpu 100%)

PR #383, opened 2025-06-20
Juan/sc 10379/list return types in the list tests output

johnwalz97:
# Pull Request Description

## What and why?

This PR adds two columns to the list test output, specifically 'has tables' and 'has figures'. These columns let users know what to expect in the test output before they run it. 

## How to test

To test, open a notebook and run this snippet of code:

```python
from validmind.tests.load import list_tests

# len(list_tests(pretty=False))
list_tests()
```

## What needs special review?

Please let me know what you think about the way in which the information is presented, specifically via flags. We could combine the information on outputs into two columns. 

There should be no breaking or logical changes besides the additional columns since the majority of changes are just return type annotations. 

## Release notes

Added columns in the `list_tests()` output to show what kind of artifacts each test in the Valid Mind library produces.

## Checklist
- [x] What and why
- [x] Screenshots or videos (Frontend)
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally
- [x] Environment variable additions/changes documented (if required)


AnilSorathiya:
@johnwalz97 @juanmleng and @cachafla 
Just wondering, how about return types of tests display in a single column where output types will be listed as values in a list form? similar to "required inputs" -> ["dataset", "model", "models", ...] 
  for example, 
  "Output types" column in a table:
  [table, figure, text,...]
  
  It will allow us to add more return types without disturbing `list_test()`.

johnwalz97:
> @johnwalz97 @juanmleng and @cachafla Just wondering, how about return types of tests display in a single column where output types will be listed as values in a list form? similar to "required inputs" -> ["dataset", "model", "models", ...] for example, "Output types" column in a table: [table, figure, text,...]
> 
> It will allow us to add more return types without disturbing `list_test()`.

That's not a bad idea. The only thing is it does make it a bit more difficult to filter the data frame down to only tests that have tables or only tests that have figures. but it would definitely make the table a bit more compact.

The thing is right now, it's set up to only be able to tell if the test outputs a table or figure. Shouldn't be that difficult to add detection for the other types. It's just a question of if we want it. Let me know what you guys think.

cachafla:
@johnwalz97 how does the user filter the list of tests for "show me tests that output figures"?

johnwalz97:
> @johnwalz97 how does the user filter the list of tests for "show me tests that output figures"?

Right now, they would have to programmatically filter the list that's returned if they don't use the pretty argument, or they would have to get the DataFrame from the Styler object and filter that. There isn't a parameter right now for filtering, but do you want to add that as part of this ticket?

cachafla (approved):
I updated `notebooks/how_to/explore_tests.ipynb`. I suggest we merge this and address improvements later (like filtering or adding something to describe_test) since this PR makes changes to every test.

PR #401, opened 2025-07-25
Support for the Deepeval dataset (LLMTestCase) for LLM tests

AnilSorathiya:
# Pull Request Description

## Checklist
- [ ] What and why
- [ ] How to test
- [ ] Labels applied
- [ ] PR linked to Shortcut
- [ ] Tested locally


validmind/row_metrics/llm/AnswerRelevancy.py
juanmleng, line 1: Perhaps not in this PR, but we may want to consider creating a dedicated namespace for external metrics (e.g., deepeval, ragas) in the near future. An alternative for now, could be to just rename this metric to `DeepEvalAnswerRelevancy`.

validmind/row_metrics/llm/AnswerRelevancy.py
```diff
@@ -0,0 +1,56 @@
+# Copyright © 2023-2024 ValidMind Inc. All rights reserved.
+# See the LICENSE file in the root of this repository for details.
+# SPDX-License-Identifier: AGPL-3.0 AND ValidMind Commercial
+
+from typing import Any, Dict
+
+from deepeval import evaluate
```
johnwalz97, line 7: Since deepeval is an optional dependency, best practice is to use a try/except block around the import here to gracefully fail.

Here is an example:
https://github.com/validmind/validmind-library/blob/main/validmind/tests/model_validation/ragas/AnswerCorrectness.py#L18-L30

cachafla (commented):
I have some suggestions to reduce the code impact/changes. Let's talk offline.

validmind/unit_metrics/classification/Accuracy.py
```diff
@@ -6,10 +6,11 @@
 
 from validmind import tags, tasks
 from validmind.vm_models import VMDataset, VMModel
+from validmind.vm_models.result.result import UnitMetricValue
```
johnwalz97, line 9: I am a fan of doing this but totally nitpicky and up to you:


```suggestion
from validmind.vm_models.result import UnitMetricValue
```

This let's you export `UnitMetricValue` from the `__init__.py` in `validmind/vm_models/result/` instead of from the file directly. That gives you the freedome to move it around without affecting imports.

validmind/unit_metrics/classification/F1.py
```diff
@@ -6,10 +6,11 @@
 
 from validmind import tags, tasks
 from validmind.vm_models import VMDataset, VMModel
+from validmind.vm_models.result.result import UnitMetricValue
```
johnwalz97, line 9: Same here and in the rest of the unit metrics and row metrics

validmind/vm_models/result/result.py
```diff
@@ -94,6 +94,223 @@ def serialize(self) -> Dict[str, Any]:
         return {key: getattr(self, key) for key in self.__dict__}
 
 
+class MetricValues(ABC):
```
johnwalz97, line 97: I would move this and the child classes to a new file since its making `result.py` pretty unwieldy

johnwalz97 (changes requested):
Couple of comments but looks good overall

validmind/unit_metrics/classification/Accuracy.py
AnilSorathiya: with new modifications, I have removed the UnitMetricValue  class.

AnilSorathiya:
1. Rename "RowLevel" metric classes to a new `Scorer` class/function
>> ✅ 
The main reason is that we don't want to treat these as `metrics` in the ValidMind sense:

- These scorers have a specific purpose of expanding/augmenting datasets with new columns
- We don't want to give the impression that the output of those scores gets saved as a "test result"
- If the user wants to save those scores then they can run a subsequent test like a `Histogram` test
- The `Scorer` notion sits closer to what we are trying to do with Ragas and DeepEval, but it also plays nicely with more basic scorers that are being added on this PR (like logloss, etc.)
- Scorers should be able to return an "object" that has more than just a score. Example: in Ragas we need to store a `score` and a `reason` for each row in the dataset
>> ✅ 

2. Example interface proposal

- Introduce a new `@scorer` decorator to identify scorers
>> ✅ 

  - Main reason is to avoid overlap/conflicts when using `load_test()` and `list_test()`
  >> currently scorer are listed in the `list_test()` >> ✅ 
This can be changed in another ticket.

  - for `load_test()`: this is necessary if we want to avoid VM running scorers as tests. I don't think this would be problematic but it's good to keep things separate?
    - for `list_test()`: we could have a separate `list_scorers()` function? This can be done at a later point
- Allow scorers to return any of: 
  - List of simple values, i.e.single score for each individual row
  - List of objects, i.e. score object, like {score, reason} for each row
>> ✅ 

3. Benefits of the above

- Avoid introducing changes to existing unit metric classes
- No need to affect existing unit metrics with `UnitMetricValue` or `RowMetricValues`
- `Scorer` functions don't need to return special types
  - We need to ensure the `OutputHandler` knows how to distinguish between scorers and metrics/tests
  - Scorer outputs don't need to saved to the backend

5. Example usage 
>> ✅ 


### Creating a scorer

```python
@scorer
def RagasHallucination(dataset):
    results = call_function()
    return [{
      score: result.score,
      reason: result.reason
    } for result in results]
```
  
**dataset before:**
| prompt | response |

### Assigning scores

```python
vm_gpt4 = vm.init_model(...)
vm_gpt5 = vm.init_model(...)

dataset.assign_scores(vm_gpt4, RagasHallucination)
dataset.assign_scores(vm_gpt5, RagasHallucination)
```

**dataset after:**
| prompt | response | gpt4_ragas_hallucination_score | gpt4_ragas_hallucination_reason | ... |

^ note that column name convention

### Running tests after the scorers populate the dataset

```python
# Histogram of hallucinations
vm.run_test("Histogram", {
  "params": {
    "columns": [...]
  }
})
```

tests/test_scorer_decorator.py
```diff
@@ -0,0 +1,552 @@
...
+
+        @_mock_scorer()
+        def auto_id_scorer(model, dataset):
+            return _MockList([2.0])
+        self.assertEqual(auto_id_scorer.scorer_id, "validmind.scorer.auto_id_scorer")
+
+        @_mock_scorer
+        def no_parens_scorer(model, dataset):
+            return _MockList([3.0])
+        self.assertEqual(no_parens_scorer.scorer_id, "validmind.scorer.no_parens_scorer")
```
cachafla, line 500: Looks like these tests in `TestScorerDecoratorSimple` are not necessary since we already have the same coverage at the top of the file?

validmind/tests/output.py
```diff
@@ -43,15 +43,9 @@ def process(self, item: Any, result: TestResult) -> None:
...
+class MetricValuesOutputHandler(OutputHandler):
     def can_handle(self, item: Any) -> bool:
-        # Accept individual numbers
-        if isinstance(item, (int, float)):
-            return True
-        # Accept lists/arrays of numbers for per-row metrics
-        if isinstance(item, (list, tuple, np.ndarray)):
-            return all(isinstance(x, (int, float, np.number)) for x in item)
-        return False
+        return isinstance(item, (int, float, list))
```
cachafla, line 48: Does this mean we can still save a list of values for a unit metric? Since the backend can't save that.

tests/test_scorer_decorator.py
AnilSorathiya: thanks. removed it.

validmind/vm_models/dataset/dataset.py
```diff
@@ -464,44 +464,44 @@ def assign_scores(
         metrics: Union[str, List[str]],
         **kwargs: Dict[str, Any],
     ) -> None:
```
juanmleng, line 466: I noticed that `assign_scores` currently requires a `VMModel` in all cases. However, some scorers may not need a `VMModel` and can work directly with the dataset, e.g., `OutlierScore`, `AnswerRelevancy`, or any LLM-based scorer, or scorers just doing data stats. Should we have `VMModel` optional?

AnilSorathiya: I am using input_id to link between vm_dataset and model. Let's see if we dont have model linked then column name will be overridden if we run the same scorer for two models using the same dataset.

PR #407, opened 2025-08-06
Support new interface assign_score to VM dataset

AnilSorathiya:
# Pull Request Description
* support new interface assign score assign
* Test for `assign_score` function
## Checklist
- [ ] What and why
- [ ] How to test
- [ ] Labels applied
- [ ] PR linked to Shortcut
- [ ] Tested locally

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -516,6 +516,301 @@ class TestTabularDataset(TestCase):
 
         self.assertIn("FunctionModel requires a callable predict_fn", str(context.exception))
 
+    def test_assign_score_single_metric(self):
+        """
+        Test assigning a single metric score to dataset
+        """
+        df = pd.DataFrame({"x1": [1, 2, 3], "x2": [4, 5, 6], "y": [0, 1, 0]})
+        vm_dataset = DataFrameDataset(
+            raw_dataset=df, target_column="y", feature_columns=["x1", "x2"]
+        )
+
+        # Train a simple model
+        model = LogisticRegression()
+        model.fit(vm_dataset.x, vm_dataset.y.ravel())
+        vm_model = init_model(input_id="test_model", model=model, __log=False)
+
+        # Assign predictions first (required for unit metrics)
+        vm_dataset.assign_predictions(model=vm_model)
+
+        # Test assign_score with single metric
+        vm_dataset.assign_score(vm_model, "F1")
+
+        # Check that the metric column was added
+        expected_column = f"{vm_model.input_id}_F1"
+        self.assertTrue(expected_column in vm_dataset.df.columns)
+
+        # Verify the column has the same value for all rows (scalar metric)
+        metric_values = vm_dataset.df[expected_column]
+        self.assertEqual(metric_values.nunique(), 1, "All rows should have the same metric value")
+
+        # Verify the value is reasonable for F1 score (between 0 and 1)
+        f1_value = metric_values.iloc[0]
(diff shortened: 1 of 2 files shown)


cachafla:
Is there an example notebook to try out the new interface?

juanmleng:
Some thoughts: 

- In my view, the primary use case for `assign_score()` is to support a specific category of test, let’s call it a _scorer_.
- A _scorer_ should operate on a per-row basis, where each row represents a single experiment or prediction instance. The _scorer_ takes this individual row as input and returns a score output for that specific case.
- While the score is typically a single numeric value (e.g., faithfulness, answer relevancy, etc), the design does not necessarily have to be constrained to scalar values, and it could in principle return a structured output (e.g., a dictionary or vector). 
- The key is that `assign_score()` should compute scores on a row-by-row basis, much like what `evaluate` in `RAGAS` or `DeepEval` does.
- Right now, all our existing tests, including the `unit_metrics`, do not behave as _scorers_. Instead, they are designed to compute global metrics over the full dataset, and rely on all rows across the dataset, to measure things like F1, accuracy, PSI, etc., which make sense only at the dataset level.

johnwalz97 (approved):
lgtm... probably want a minor version bump for this one since its a new interface?

notebooks/how_to/assign_score_complete_tutorial.ipynb
cachafla, line 1: Suggested notebook fix:

Remove this extra text in `### Multi-Model Scoring` that is not doing anything:

```
# Create a comparison summary
print("\n Model Performance Comparison:")
```

cachafla (approved):
Looks good to me but I'd suggest renaming the interface to `assign_scores()`. Two main reasons:

- To stay consistent with our other `assign_predictions()` interface
- To acknowledge that you can assign 1 or more scores to a dataset, not just 1

cachafla:
Great notebook 😄. I'm excited about the possibility of using scorers in the context of Gen AI use cases.

juanmleng:
I think this interface looks great. A couple of suggestions: 

- A fantastic side effect of `assign_scores()` is that it leverages `unit_metrics` which were in a kind of _waiting-to-be-used_ mode. So this is great. I like the idea of clearly defining which types of tests are supported, as it provides guidance and consistency, and I think this is a nice approach for VM-managed metrics.

- We know users will likely want to add their own scorers, so in future iterations we may want to support custom tests. For now, we could include a note in the notebook making it clear that only unit metrics are supported, so users are aware.

- The notebook is quite comprehensive and clearly shows everything you can do with `assign_scores()`. I think it would be the icing on the cake if it wrapped up with two or three of the generalized tests you implemented recently. That would make it even clearer how cool `assign_scores()` can be! 😊

juanmleng (approved):
This looks quite cool. Great job! I just left a couple of minor suggestions.

notebooks/how_to/assign_score_complete_tutorial.ipynb
AnilSorathiya: done thanks

AnilSorathiya:
> Looks good to me but I'd suggest renaming the interface to `assign_scores()`. Two main reasons:
> 
> * To stay consistent with our other `assign_predictions()` interface
> * To acknowledge that you can assign 1 or more scores to a dataset, not just 1

Make sense. Renamed to `assign_scores`.

AnilSorathiya:
> * only unit metrics are supported
* Updated the notebook description by adding this feature supports on unit_metrics only
 * The `Individual Metrics` section includes new tests

PR #418, opened 2025-08-26
Hotfix: Remove LLM is_configured() Check from Test Description Generation

johnwalz97:
# Pull Request Description

## What and why?

We no longer need the is_configured() check in the test description generation flow as we call the ValidMind API to generate descriptions. Removing as it will cause test description generation to fail in environments where an LLM provider is not configured.

## Checklist
- [ ] What and why
- [ ] How to test
- [ ] Labels applied
- [ ] PR linked to Shortcut
- [ ] Tested locally


cachafla (approved):
Nice

PR #505, opened 2026-05-13
[SC-16155] Fix async chart uploads by pre-serializing figure images

juanmleng:
# Pull Request Description

## What and why?
Tests that generate charts or figures fail to upload results to ValidMind with: 
```
RuntimeError: Timeout context manager should be used inside a task
```
and the customer confirmed that after the `Kaleido 1.2.0` fix the process no longer hangs indefinitely but still fails to complete chart generation and upload the results.

This change pre-serializes figure PNGs before `TestResult.log()` enters the async upload flow, avoiding the Plotly/Kaleido and asyncio conflict in the customer-reported error path. It also centralizes figure PNG serialization behind a cached path so uploads, HTML rendering, and base64 generation all reuse the same image bytes consistently.

## How to test
I was not able to reproduce this locally. I tested with `Python 3.9` and also tried reproducing the issue outside ValidMind by serializing figures directly through the `Figure` object, but could not trigger the failure, so this may depend on the customer’s environment. 

## What needs special review?
Check whether it makes sense to take a defensive approach here, based on the customer-provided evidence, which includes clear stack traces showing the exact error path this fix addresses.

The fix is in principle low risk because it does not change behavior on systems where serialization already works; it only guarantees that `to_image()` is not called from within an `async` context where it may conflict.

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [ ] Tested locally


validmind/vm_models/figure.py
```diff
@@ -119,58 +139,21 @@ def _get_b64_url(self):
...
-        )
+        png_bytes = self._get_png_bytes()
+        b64_data = base64.b64encode(png_bytes).decode("utf-8")
+        return f"data:image/png;base64,{b64_data}"
 
     def serialize_files(self):
         """Creates a `requests`-compatible files object to be sent to the API."""
-        if is_matplotlib_figure(self.figure):
-            buffer = BytesIO()
-            self.figure.savefig(buffer, bbox_inches="tight")
```
johnwalz97, line 150: just fyi the bbox_inches param isn't maintained in the new implementation

validmind/vm_models/figure.py
```diff
@@ -71,39 +72,58 @@ def __post_init__(self):
...
+        otherwise it generates them on demand. Caching is important for Plotly
+        figures because to_image() uses kaleido which conflicts with asyncio
+        event loops when called from within an async context.
+        """
+        if self._cached_png_bytes is not None:
+            return self._cached_png_bytes
+
+        if is_matplotlib_figure(self.figure):
+            buffer = BytesIO()
+            self.figure.savefig(buffer, format="png")
```
johnwalz97, line 88: we might want the bbox_inches (see below) but not sure

johnwalz97 (approved):
lgtm... was also not able to reproduce but it seems like this could be a fix

validmind/vm_models/figure.py
juanmleng: Great catch @johnwalz97! I am inclined to not introduced too many more changes on this PR and perhaps leave it as-is for now. Previously `bbox_inches="tight"`only affected uploads, but with the new shared PNG path it would now affect all Matplotlib rendering, not just upload serialization. If we decide we want that back, we can always reintroduce it later?

cachafla (approved):

PR #514, opened 2026-05-19
[SC-15023] feat(figures): add title field for figures and route to caption registry

johnwalz97:
## What and why?
Adds a `title` field to `Figure` so test authors can caption charts/figures the same way they already can for tables. Closes the gap between regular text blocks (where CKEditor figures already support captions) and test-driven blocks (where the validmind library had no way to set one).

Test functions can now return `{"My Chart Title": fig, ...}` — mirroring the existing dict-of-tables convention — and the title flows through to the platform's document media registry as `Figure N. <title>`.

Also fixes a quiet bug for tables: `ResultTable.title` was serialized as `metadata.title`, but the backend caption registry reads `metadata.caption`. We now emit both keys, so table titles actually reach the registry.

**Before:** test-driven figures had no caption-registry entry; tables had a `title` but it was only rendered as an inline `<h4>` heading and never reached the "Table N. <caption>" auto-numbered caption.
**After:** both figures and tables flow titles into the registry and get auto-numbered captions in the doc.

No backend, frontend, or CKEditor changes are required — that plumbing already reads `metadata.caption` and renders captions via the existing `DocumentMediaCaption` component.

## How to test
1. `pytest tests/test_results.py` — 4 new tests cover the new serialization + dict-of-figures handler (28/28 pass locally).
2. In a notebook, define a custom test that returns `{"My Cool Chart": some_plotly_fig}` and log the result against a model with the `global_media_captions_enabled` org setting on. The figure should render with `Figure N. My Cool Chart` underneath it.
3. The updated example in `notebooks/how_to/tests/custom_tests/implement_custom_tests.ipynb` (the `ComplexOutput` test) now showcases titled figures end-to-end.

## What needs special review?
- `ResultTable.serialize()` now emits both `title` and `caption` in metadata. Worth confirming no downstream consumer (e.g. docs export, PDF generation) chokes on the extra key. None found in backend code search, but flagging for visibility.
- `FigureOutputHandler.can_handle` now claims dicts whose values are all figure-like. It's ordered before `TableOutputHandler` in the handler list, so dict-of-DataFrames continues to be handled by the table handler.

- No breaking changes — `title` is a new optional field on `Figure` with a default; the dict-of-figures syntax is additive.
- No DB migrations.
- No backend/frontend deploys required to ship this; benefits appear automatically once tenants have `global_media_captions_enabled`.

## Release notes
Custom tests can now title figures the same way they already title tables. Return `{"My Chart Title": fig}` from a test (or pass `title="…"` to `Figure`/`ResultTable`) and the title will be rendered as `Figure N. My Chart Title` in the document's caption registry.

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut (via sc-15023 branch prefix)
- [x] Unit tests added
- [x] Tested locally
- [x] Documentation updated (docstrings + example notebook)
- [ ] Environment variable additions/changes documented — N/A

### DIFF
--- a/notebooks/how_to/tests/custom_tests/implement_custom_tests.ipynb
+++ b/notebooks/how_to/tests/custom_tests/implement_custom_tests.ipynb
@@ -92,7 +92,7 @@
     "\n",
     "### Key concepts\n",
     "\n",
-    "**Model documentation**: A structured and detailed record pertaining to a model, encompassing key components such as its underlying assumptions, methodologies, data sources, inputs, performance metrics, evaluations, limitations, and intended uses. It serves to ensure transparency, adherence to regulatory requirements, and a clear understanding of potential risks associated with the model’s application.\n",
+    "**Model documentation**: A structured and detailed record pertaining to a model, encompassing key components such as its underlying assumptions, methodologies, data sources, inputs, performance metrics, evaluations, limitations, and intended uses. It serves to ensure transparency, adherence to regulatory requirements, and a clear understanding of potential risks associated with the model\u2019s application.\n",
     "\n",
     "**Documentation template**: Functions as a test suite and lays out the structure of model documentation, segmented into various sections and sub-sections. Documentation templates define the structure of your model documentation, specifying the tests that should be run, and how the results should be displayed.\n",
     "\n",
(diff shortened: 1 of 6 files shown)


juanmleng (approved):
Really cool @johnwalz97!

PR #526, opened 2026-06-26
[SC 16935] Fix IQROutliersBarPlot failure on boolean numeric

AnilSorathiya:
# Pull Request Description

## What and why?

`IQROutliersBarPlot` failed with `TypeError: numpy boolean subtract` when a dataset included boolean columns in `feature_columns_numeric` (pandas treats `bool` as numeric).

**Before:** Boolean/binary columns could still be processed when building `outlier_counts_by_feature`, causing `quantile()` to fail on boolean data.

**After:** Boolean and binary features are excluded from IQR outlier calculations using `is_bool_dtype()` and `nunique() > 2`, for both plots and raw data output.

## How to test

```bash
pytest tests/unit_tests/data_validation/test_IQROutliersBarPlot.py -v
```

Or manually:

```python
import numpy as np
import pandas as pd
import validmind as vm
from validmind.tests.data_validation.IQROutliersBarPlot import IQROutliersBarPlot

df = pd.DataFrame({
    "numeric": np.random.randn(100),
    "flag": np.random.choice([True, False], 100),
})
dataset = vm.init_dataset(input_id="repro", dataset=df, __log=False)
IQROutliersBarPlot(dataset)  # should complete without error
```

## What needs special review?

- Confirm excluding boolean columns from IQR analysis is the intended behavior (IQR is not meaningful for boolean features).

- No dependencies.
- No breaking changes.
- Boolean columns will no longer appear in `IQROutliersBarPlot` output.

## Release notes

`bug`

Fixed a failure in `IQROutliersBarPlot` when datasets contain boolean feature columns. The test now skips boolean and binary features instead of raising a `numpy boolean subtract` error during IQR calculations.

## Checklist
- [x] What and why
- [x] How to test
- [ ] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally

### DIFF
--- a/tests/unit_tests/data_validation/test_IQROutliersBarPlot.py
+++ b/tests/unit_tests/data_validation/test_IQROutliersBarPlot.py
@@ -97,4 +97,4 @@ class TestIQROutliersBarPlot(unittest.TestCase):
         results = IQROutliersBarPlot(vm_dataset)
         raw_data = results[-1]
 
-        self.assertNotIn("flag", raw_data.outlier_counts_by_feature.index)
+        self.assertNotIn("flag", raw_data.outlier_counts_by_feature.columns)
--- a/tests/unit_tests/data_validation/test_IQROutliersTable.py
+++ b/tests/unit_tests/data_validation/test_IQROutliersTable.py
@@ -83,3 +83,25 @@ class TestIQROutliersTable(unittest.TestCase):
         # Verify binary column is not in results
         for summary in outliers_summary:
             self.assertNotIn("binary", summary["Variable"])
+
+    def test_boolean_dtype_excluded_from_raw_data(self):
+        n_samples = 100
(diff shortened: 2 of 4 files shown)


juanmleng:
Nice fix overall — the `is_bool_dtype` guard is the right call and the intent is much clearer than the old `unique() > 2` alone.

Two things worth addressing before merge:

- **The new test is checking the wrong attribute.** `test_boolean_dtype_excluded_from_raw_data` asserts `"flag" not in raw_data.outlier_counts_by_feature.index`, but `outlier_counts_by_feature` is a DataFrame — `.index` holds row integers, not column names. So this test always passes regardless of whether the fix is working. It should be `.columns` instead.

- **`IQROutliersTable` has the same bug.** It uses `len(df[col].unique()) <= 2` without the `is_bool_dtype` guard, so a bool-dtype column would hit the same crash there. Worth patching in the same PR for consistency.

juanmleng (approved):

PR #527, opened 2026-06-26
[SC 16940] Fix matplotlib colormap API in statsmodels histogram tests

AnilSorathiya:
# Pull Request Description

## What and why?
Replaces deprecated `matplotlib.cm.get_cmap("viridis")` with `matplotlib.colormaps["viridis"]` in three statsmodels visualization tests:

- `PredictionProbabilitiesHistogram`
- `ScorecardHistogram`
- `CumulativePredictionProbabilities`

**Before:** Tests failed on newer matplotlib versions with `AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'`.

**After:** Colormaps are resolved via the supported matplotlib API, and the histogram tests run successfully across Python 3.9-3.14.

### Additional change: `ModelPredictionResiduals` KS test

This PR also updates the Kolmogorov-Smirnov normality test in `ModelPredictionResiduals` from the string-based scipy call:

```python
kstest(residuals, "norm", args=(residuals.mean(), residuals.std()))
```

to passing an explicit CDF callable:

```python
kstest(residuals, norm(loc=residuals.mean(), scale=residuals.std()).cdf)
```

This is equivalent for the current scipy API and produced the same statistic and p-value in local verification. It is not required for the matplotlib fix; it is included as a small cleanup to make the test more explicit and align with scipy's preferred callable-style usage.

### Dependency fix for python 3.14 build

The failure comes from **`shap`** (in the `[all]` extra), not `scorecardpy`.

On Linux CI, `shap` declares `numba` and `llvmlite` without a minimum version. The resolver then picks the oldest compatible pair:

- `numba==0.53.1`
- `llvmlite==0.36.0`

Those versions only support Python up to 3.9/3.10, so install fails on Python 3.14.

Explicit Python 3.14 floors were added in `pyproject.toml` for the extras that pull in `shap`:

```toml
"numba (>=0.65.1) ; python_version >= '3.14'",
"llvmlite (>=0.47.0,<0.48) ; python_version >= '3.14'",
```

Older Python versions (3.9–3.13) are unchanged. 

## How to test
Run the affected unit tests locally:

```bash
python -m unittest tests.unit_tests.model_validation.statsmodels.test_PredictionProbabilitiesHistogram
python -m unittest tests.unit_tests.model_validation.statsmodels.test_ScorecardHistogram
```

Or run the full statsmodels unit test suite if preferred.

## What needs special review?
Please review the dependency notes alongside the `ModelPredictionResiduals` cleanup. The `kstest` change is intentional and behavior-preserving; the scipy version constraint should remain aligned with the intended compatibility policy before merge.

- No new dependencies
- No breaking changes
- Minor scipy API cleanup in `ModelPredictionResiduals`
- The scipy version constraint should be reviewed separately from the matplotlib API fix so it does not look accidental

## Release notes
Fixes statsmodels histogram tests that failed on newer matplotlib versions due to removal of the deprecated `cm.get_cmap()` API.

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally


juanmleng:
The matplotlib fix is correct — `matplotlib.colormaps["viridis"]` is the right replacement for the removed `cm.get_cmap()` API, and it's applied consistently across all three affected files. Clean.

One thing worth discussing before merging:

**There's a quiet undocumented change in `ModelPredictionResiduals.py`** that's a bit confusing alongside the scipy pin. The PR switches `kstest` from the old string-based call `kstest(residuals, "norm", args=(...))` to `kstest(residuals, norm(...).cdf)` — which is correct and forward-compatible with scipy 2.0. But `pyproject.toml` simultaneously pins `scipy <2.0`. These two things pull in opposite directions: the code is updated as if scipy 2.0 is coming, but the pin keeps it out. Worth clarifying the intent — if scipy 2.0 has other unresolved issues in the library, the pin makes sense but the description should say so. If it doesn't, the `<2.0` cap should probably be dropped.

The `kstest` change itself is fine and the new call is actually cleaner. Just needs a note in the PR description so it doesn't look accidental.

johnwalz97 (approved):
lgtm

juanmleng (approved):

AnilSorathiya:
> This is equivalent for the current scipy API and produced the same statistic and p-value in local verification. It is not required for the matplotlib fix; it is included as a small cleanup to make the test more explicit and align with scipy's preferred callable-style usage.

Removed scipy <2.0 dependency. 
This is equivalent for the current scipy API and produced the same statistic and p-value in local verification. It is not required for the matplotlib fix; it is included as a small cleanup to make the test more explicit and align with scipy's preferred callable-style usage.

PR #528, opened 2026-06-26
[SC-16871] Make DeepEval agentic tests natively accessible within the default ValidMind test suite for Enterprise Google Gemini (Vertex AI)

juanmleng:
# Pull Request Description

## What and why?

Before this change, calling `set_judge_config()` to configure a judge LLM applied to
prompt-validation and RAGAS tests but was silently ignored by DeepEval scorers. Scorers
in `validmind.scorers.llm.deepeval` had their own provider-detection logic in
`get_deepeval_model()` that never consulted the configured judge, so enterprise customers
using Vertex AI (or any non-default provider) had to write custom scorer boilerplate to
work around it.

This PR wires `get_deepeval_model()` to honour `set_judge_config()`: when a judge LLM
has been explicitly configured, all five built-in DeepEval scorers (`AnswerRelevancy`,
`ArgumentCorrectness`, `Hallucination`, `PlanAdherence`, `PlanQuality`, `ToolCorrectness`)
now use it automatically — including Vertex AI via `langchain-google-vertexai`.

Internally, the provider-specific `GeminiDeepEvalModel` wrapper is replaced by a generic
`LangChainDeepEvalModel` that wraps any `BaseChatModel`, and `_build_gemini_deepeval_model`
is simplified to delegate to it.

The `configure_judge_llms.ipynb` notebook is updated to document this behaviour and
includes an optional `set_judge_config()` example cell.

## How to test

**Automated tests**

```bash
cd validmind-library
make test-unit
```

All 89 test suites pass with no changes to existing tests.

**Manual testing**

```python
from langchain_google_genai import ChatGoogleGenerativeAI, GoogleGenerativeAIEmbeddings
from validmind.ai.utils import set_judge_config, get_deepeval_model

set_judge_config(
    judge_llm=ChatGoogleGenerativeAI(model="gemini-3.5-flash"),
    judge_embeddings=GoogleGenerativeAIEmbeddings(model="models/text-embedding-004"),
)

model = get_deepeval_model()
assert type(model).__name__ == "LangChainDeepEvalModel"
assert model.get_model_name() == "gemini-3.5-flash"
```

Then run a built-in scorer end-to-end:

```python
import pandas as pd, validmind as vm

df = pd.DataFrame({
    "input": ["What is the capital of France?"],
    "actual_output": ["Paris."],
    "context": [["France's capital is Paris."]],
})
ds = vm.init_dataset(dataset=df, input_id="test", text_column="input", target_column="actual_output")
ds.assign_scores(metrics=["validmind.scorers.llm.deepeval.AnswerRelevancy"])
```

Expected: DataFrame with a non-null score column, no errors.

## What needs special review?

- **`get_deepeval_model()` now checks `__judge_llm` before provider detection.** This is
  the intended priority order, but reviewers should confirm there are no edge cases where
  a stale `__judge_llm` from a previous `set_judge_config()` call in the same session
  would produce unexpected results in a notebook that later switches providers via env vars.
- **`get_model_name()` introspection** — the new `LangChainDeepEvalModel` checks `model`,
  `model_name`, and `azure_deployment` attributes in order. If a custom `BaseChatModel`
  exposes none of these, it falls back to the class name. Confirm this is acceptable for
  all supported providers.

None. No new dependencies, no migrations, no env var changes. The existing Gemini
deepeval path (`_build_gemini_deepeval_model`) is preserved and now delegates to the
new generic wrapper.

No companion PRs in frontend or documentation repos.

## Release notes

This is an `enhancement`. Suggested label: `enhancement`.

DeepEval scorers (`AnswerRelevancy`, `Hallucination`, and others in
`validmind.scorers.llm.deepeval`) now respect the judge LLM configured via
`set_judge_config()`. This means any provider — including Vertex AI — configured once
for prompt-validation and RAGAS tests is automatically used by scorers as well, with no
additional setup required.

## Checklist
- [x] What and why
- [x] How to test
- [ ] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally

---

**Shortcut:** [SC-16871](https://app.shortcut.com/validmind/story/16871)

### DIFF
--- a/validmind/ai/utils.py
+++ b/validmind/ai/utils.py
@@ -238,14 +238,17 @@ def _unwrap_deepeval_response(response):
     return getattr(response, "content", response)
 
 
-def _build_gemini_deepeval_model(model):
-    DeepEvalBaseLLM = _import_deepeval_base_llm()
-    judge_llm, _ = _build_gemini_judge_config(model)
+def _build_langchain_deepeval_model(chat_model):
+    """Wrap any LangChain BaseChatModel as a DeepEval-compatible model.
 
-    class GeminiDeepEvalModel(DeepEvalBaseLLM):
-        def __init__(self, chat_model, model_name):
-            self._chat_model = chat_model
-            self._model_name = model_name
+    Used when set_judge_config() has been called so that DeepEval scorers
+    honour the same judge model as prompt-validation and RAGAS tests.
+    """
+    DeepEvalBaseLLM = _import_deepeval_base_llm()
+
+    class LangChainDeepEvalModel(DeepEvalBaseLLM):
+        def __init__(self, model):
+            self._chat_model = model
             self.model = self.load_model()
 
         def load_model(self, *args, **kwargs):
@@ -257,7 +260,6 @@ def _build_gemini_deepeval_model(model):
                 response = chat_model.with_structured_output(schema).invoke(prompt)
             else:
                 response = chat_model.invoke(prompt)
-
             return _unwrap_deepeval_response(response)
 
         async def a_generate(self, prompt: str, schema=None):
@@ -268,13 +270,21 @@ def _build_gemini_deepeval_model(model):
                 )
             else:
                 response = await chat_model.ainvoke(prompt)
-
             return _unwrap_deepeval_response(response)
 
         def get_model_name(self, *args, **kwargs):
-            return self._model_name
+            for attr in ("model", "model_name", "azure_deployment"):
+                val = getattr(self._chat_model, attr, None)
+                if val:
+                    return val
+            return type(self._chat_model).__name__
 
-    return GeminiDeepEvalModel(judge_llm, model)
+    return LangChainDeepEvalModel(chat_model)
+
+
+def _build_gemini_deepeval_model(model):
+    judge_llm, _ = _build_gemini_judge_config(model)
+    return _build_langchain_deepeval_model(judge_llm)
 
 
 def run_deepeval_evaluation(*, test_cases, metrics):
@@ -332,6 +342,9 @@ def get_deepeval_model():
     OpenAI/Azure scorers currently pass a model string. Gemini support requires a
     native DeepEval model object so the provider can be configured correctly.
     """
+    global __judge_llm
+    if __judge_llm is not None:
+        return _build_langchain_deepeval_model(__judge_llm)
 
     provider = _get_configured_provider()
     _, model = get_client_and_model()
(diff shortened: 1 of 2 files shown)


validmind/ai/utils.py
```diff
@@ -332,6 +342,9 @@ def get_deepeval_model():
     OpenAI/Azure scorers currently pass a model string. Gemini support requires a
     native DeepEval model object so the provider can be configured correctly.
     """
+    global __judge_llm
+    if __judge_llm is not None:
+        return _build_langchain_deepeval_model(__judge_llm)
```
johnwalz97: Heads up: gating on `if __judge_llm is not None` triggers on more than `set_judge_config()`. `__judge_llm` also gets set as a side effect of `get_judge_config()` (line 334), which runs on every prompt-validation test, every RAGAS test, and `is_configured()`. So once any of those run in a session, this branch is taken for everything after.

Checked against `deepeval 3.4.0` with `OPENAI_API_KEY` set:
```
scorer first        -> get_deepeval_model() = "gpt-4.1"  -> GPTModel               (using_native_model=True)
scorer after a test -> get_deepeval_model() = LangChainDeepEvalModel  (using_native_model=False)
```
Same env, same scorer, different eval path depending on what ran first in the kernel.

Probably want to key off intent instead of the cached global:
```python
# module level
__judge_llm_explicitly_set = False

# in set_judge_config(), after validation:
global __judge_llm_explicitly_set
__judge_llm_explicitly_set = True

# here:
if __judge_llm_explicitly_set and __judge_llm is not None:
    return _build_langchain_deepeval_model(__judge_llm)
```
Also evens out a small inconsistency: this checks `is not None` while `get_judge_config` (line 325) and `set_judge_config` both require embeddings too.

johnwalz97 (approved):
lgtm but claude raised a finding that i think needs addressed

AnilSorathiya (approved):
Looking good. Points to consider:

1. Embeddings required even when DeepEval doesn’t use them

get_deepeval_model() requires both __judge_llm and __judge_embeddings to be set. DeepEval scorers don’t use embeddings, but callers must still pass an embeddings object to set_judge_config(). We might keep embeddings are optional.

2. No per-scorer judge override (Could be separate work)

Prompt-validation and RAGAS support per-call judge_llm/judge_embeddings overrides. All 14 DeepEval scorers call get_deepeval_model() with no parameters — global config only.

3. No reset API

Once set_judge_config() is called, there’s no reset_judge_config() to clear __judge_llm_explicitly_set. Session restart is the only way back to env-based defaults.

cachafla (approved):

PR #529, opened 2026-06-29
fix: MinimumF1Score multiclass detection uses y_true and y_pred

hunner:
# Pull Request Description

## What and why?

`MinimumF1Score` raised `ValueError: Target is multiclass but average='binary'` for multiclass
models (ZD-704).

The test chose binary-vs-macro averaging from `len(np.unique(dataset.y))` — the **true labels only**.
scikit-learn's `f1_score` determines the target type from the union of `y_true` *and* `y_pred`. When a
split's true labels collapse to ≤2 distinct classes while the model predicts a third class, the binary
branch is chosen and sklearn rejects it. This is easy to trigger with an `input_grid` over
`[train, test]`: a 3-class model's train split usually contains all classes (macro branch, fine), but
the smaller test/holdout split can contain only 2 of the 3 true classes while the model still predicts
the third — so the comparison passes on train and crashes on test.

Before: `f1_score(y_true, y_pred)` (binary default) on `y_true=[0,1,1,0,1]`, `y_pred=[0,1,2,0,1]` →
`ValueError`. After: the union of labels is `{0,1,2}` → macro averaging → `0.6`.

The invariant restored: the averaging mode must match the label space `f1_score` actually evaluates,
which includes predicted labels, not just ground-truth labels.

This PR also adds a one-line note to `ROCCurve` and `PrecisionRecallCurve`, which use the same
`len(np.unique(y_true)) > 2` text but intentionally **skip** multiclass (they compute on
probabilities), so the guard there is not mistakenly "fixed" to inspect `y_pred`.

## How to test

```python
# any classifier whose predictions can include a class absent from a given split's true labels,
# e.g. a 3-class model evaluated over input_grid=[train_ds, test_ds]
run_test(
    "validmind.model_validation.sklearn.MinimumF1Score",
    input_grid={"dataset": [vm_train_ds, vm_test_ds], "model": [vm_logistic_model]},
    params={"min_threshold": 0.5},
).log()
```

New unit tests in `tests/unit_tests/model_validation/sklearn/test_MinimumF1Score.py` cover both a
binary case (binary averaging, unchanged) and a multiclass case where predictions include a class
absent from `y_true` (macro averaging, no crash), asserting the score equals sklearn's expected value
so the correct branch is provably taken.

## What needs special review?

The main review surface is the class-count computation, which now unions `y_true` and `y_pred` before
deciding the averaging mode. For genuinely binary problems (both true and predicted within two classes)
behaviour is unchanged — only previously-crashing paths switch to macro, consistent with the test's
documented "for multi-class problems, macro averaging is used."

None. No API or signature changes; no new dependencies.

## Release notes

`bug` — Fixed `MinimumF1Score` raising "Target is multiclass but average='binary'" when a model
predicts a class not present in a dataset split's true labels.

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally

Closes ZD #704 / [sc-16970](https://app.shortcut.com/validmind/story/16970)

### DIFF
--- a/validmind/tests/model_validation/sklearn/MinimumF1Score.py
+++ b/validmind/tests/model_validation/sklearn/MinimumF1Score.py
@@ -58,10 +58,19 @@ def MinimumF1Score(
     closely with specific requirements.
     """
 
-    if len(np.unique(dataset.y)) > 2:
-        score = f1_score(dataset.y, dataset.y_pred(model), average="macro")
+    y_true = dataset.y
+    y_pred = dataset.y_pred(model)
+
+    # Decide the averaging method from the labels f1_score actually sees -- the
+    # union of y_true and y_pred. Inspecting y_true alone is not enough: a split
+    # whose true labels collapse to <=2 classes while the model predicts a third
+    # class is still multiclass to sklearn, and average="binary" would raise
+    # "Target is multiclass but average='binary'".
+    n_classes = len(np.unique(np.concatenate([y_true, y_pred])))
+    if n_classes > 2:
+        score = f1_score(y_true, y_pred, average="macro")
     else:
-        score = f1_score(dataset.y, dataset.y_pred(model))
+        score = f1_score(y_true, y_pred)
 
     return (
         [
--- a/tests/unit_tests/model_validation/sklearn/test_MinimumF1Score.py
+++ b/tests/unit_tests/model_validation/sklearn/test_MinimumF1Score.py
@@ -1,0 +1,76 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import f1_score
+
+import validmind as vm
+from validmind.tests.model_validation.sklearn.MinimumF1Score import MinimumF1Score
+
+
+def _dataset_with_predictions(input_id, y_true, y_pred):
+    """Build a VMDataset whose predictions are injected verbatim.
+
+    A fitted model is only needed to construct a valid VMModel; the predictions
+    are supplied via ``prediction_values`` so the model's ``predict`` is never
+    called and the true/predicted label sets can be controlled exactly.
+    """
+    df = pd.DataFrame(
+        {
+            "f1": np.linspace(-1.0, 1.0, len(y_true)),
+            "f2": np.linspace(1.0, -1.0, len(y_true)),
+            "target": y_true,
+        }
+    )
+    dataset = vm.init_dataset(
+        input_id=input_id, dataset=df, target_column="target", __log=False
+    )
+
+    model = LogisticRegression(max_iter=1000)
+    model.fit(df[["f1", "f2"]].to_numpy(), np.array(y_true))
+    vm_model = vm.init_model(input_id=f"{input_id}_model", model=model, __log=False)
+
+    dataset.assign_predictions(model=vm_model, prediction_values=y_pred)
+    return dataset, vm_model
+
+
+class TestMinimumF1Score(unittest.TestCase):
+    def test_predicted_class_absent_from_true_labels(self):
+        # Regression test for ZD-704. This split's true labels are binary ({0, 1}),
+        # but the model predicts a third class. scikit-learn's f1_score derives the
+        # target type from the union of y_true and y_pred, so deciding the averaging
+        # mode from y_true alone selects average="binary" and raises
+        # "Target is multiclass but average='binary'". The test must instead detect
+        # the multiclass label space and use macro averaging.
+        y_true = [0, 1, 1, 0, 1, 0, 1, 0]
+        y_pred = [0, 1, 2, 0, 1, 0, 2, 0]  # class 2 never appears in y_true
+        dataset, model = _dataset_with_predictions("f1_multiclass_pred", y_true, y_pred)
+
+        result = MinimumF1Score(dataset, model, min_threshold=0.5)
+
+        score = result[0][0]["Score"]
+        expected = f1_score(np.array(y_true), np.array(y_pred), average="macro")
+        # Matching the macro value confirms the multiclass branch was taken (the
+        # binary branch would have raised rather than returned a number).
+        self.assertAlmostEqual(score, expected)
+        self.assertEqual(result[1], score > 0.5)
+
+    def test_binary_uses_binary_average(self):
+        # A genuinely binary problem (true and predicted labels both within {0, 1})
+        # must still use sklearn's default binary averaging and be unaffected by the
+        # fix.
+        y_true = [0, 1, 1, 0, 1, 0]
+        y_pred = [0, 1, 0, 0, 1, 1]
+        dataset, model = _dataset_with_predictions("f1_binary", y_true, y_pred)
+
+        result = MinimumF1Score(dataset, model, min_threshold=0.5)
+
+        score = result[0][0]["Score"]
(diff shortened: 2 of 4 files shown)


tests/unit_tests/model_validation/sklearn/test_MinimumROCAUCScore.py
```diff
@@ -0,0 +1,73 @@
...
+        # class) column for class 3 — the macro average must skip such columns so
+        # roc_auc_score does not raise "Only one class present in y_true".
+        y_true = [0, 1, 2, 0, 1, 2, 0, 1, 2]
+        y_pred = [0, 1, 3, 0, 1, 2, 0, 2, 2]  # class 3 never appears in y_true
+        dataset, model = _dataset_with_predictions("roc_multiclass_pred", y_true, y_pred)
+
+        result = MinimumROCAUCScore(dataset, model, min_threshold=0.5)
+
+        score = result[0][0]["Score"]
+        self.assertTrue(0.0 <= score <= 1.0)
```
juanmleng: The F1Score tests check that the returned score exactly matches what sklearn would compute — which is the right way to do it, because if the wrong branch is taken the number will be different, not just invalid. The ROC AUC tests only check that `0.0 <= score <= 1.0`. That would pass even if the score were completely wrong. For the common-case test especially (where predictions stay within the true classes), you can compute the expected value directly from numpy/sklearn and assert equality, same as the F1 tests do. Worth fixing so this test actually catches a regression.

validmind/tests/model_validation/sklearn/MinimumROCAUCScore.py
```diff
@@ -63,15 +63,27 @@ def MinimumROCAUCScore(
...
 
         y_true_binary = lb.transform(y_true)
-        y_score_binary = lb.transform(dataset.y_pred(model))
+        y_score_binary = lb.transform(y_pred)
+
+        # ROC AUC is only defined for classes that have both positive and negative
+        # examples in y_true; restrict the macro average to those columns so a
+        # prediction-only class does not raise "Only one class present in y_true".
+        n_samples = y_true_binary.shape[0]
+        scorable = (y_true_binary.sum(axis=0) > 0) & (y_true_binary.sum(axis=0) < n_samples)
```
juanmleng: After the fix, the code filters down to only the classes that have both positive and negative examples in the true labels — which is the right thing to do. But if that filter leaves zero classes (imagine a split so small or lopsided that every class either appears in every row or none at all), the code passes an empty array to sklearn and gets an unhelpful error. This is an edge case, but other tests in the repo (like `ROCCurve.py`) handle this kind of situation by raising a `SkipTestError` with a clear message. Worth doing the same here so the user gets a readable error instead of a crash from deep inside sklearn.

validmind/tests/model_validation/sklearn/MinimumF1Score.py
```diff
@@ -58,10 +58,19 @@ def MinimumF1Score(
...
-        score = f1_score(dataset.y, dataset.y_pred(model), average="macro")
+    y_true = dataset.y
+    y_pred = dataset.y_pred(model)
+
+    # Decide the averaging method from the labels f1_score actually sees -- the
+    # union of y_true and y_pred. Inspecting y_true alone is not enough: a split
+    # whose true labels collapse to <=2 classes while the model predicts a third
+    # class is still multiclass to sklearn, and average="binary" would raise
+    # "Target is multiclass but average='binary'".
+    n_classes = len(np.unique(np.concatenate([np.unique(y_true), np.unique(y_pred)])))
```
juanmleng: The code currently deduplicates `y_true` and `y_pred` individually before combining them, then deduplicates again on the combined result. The middle step doesn't do anything — you get the same answer whether or not you deduplicate before combining. Not wrong, just noisier than it needs to be. One `np.unique` on the concatenated arrays is enough. (Same pattern in `MinimumROCAUCScore.py`.)

juanmleng:
**Follow-ups (non-blocking):**

- The `MinimumROCAUCScore` docstring doesn't mention that when a model predicts a class that's absent from the split's true labels, that class gets excluded from the score. A one-liner would help callers understand what the number actually represents in that situation.
- The helper `_dataset_with_predictions` is copy-pasted identically into both new test files. Fine for now, but worth moving to a shared utility once more sklearn tests follow this pattern.
- `ROCCurve.py` and `PrecisionRecallCurve.py` have the same-looking `len(np.unique(y_true)) > 2` guard but intentionally skip multiclass rather than handling it. A short comment there pointing out that this is deliberate would prevent someone from "fixing" it later based on this PR.

juanmleng (approved):
Thanks @hunner! I left few minor comments. Other than that, lgtm!

tests/unit_tests/model_validation/sklearn/test_MinimumROCAUCScore.py
hunner: Good catch — and digging into this changed the outcome. I verified the `MinimumROCAUCScore` change was a **no-op**: it produced identical scores to the original in every case, including the predicted-class-absent edge:

```
common (pred subset)          old=0.916667  new=0.916667
edge  (pred class 3 absent)   old=0.861111  new=0.861111
```

The original already represents an out-of-vocabulary prediction as an all-negative indicator row (which penalizes it correctly), and in the multiclass branch every true-class column always has both a positive and a negative, so it never crashes. Rather than strengthen a test for a change that changes nothing, I reverted the `MinimumROCAUCScore` change and removed this test. Only the genuine `MinimumF1Score` fix remains. (48a05d8b)

🤖 _Automated reply_

validmind/tests/model_validation/sklearn/MinimumROCAUCScore.py
hunner: Reverted — see the note on the test thread. This empty-filter edge only existed *because* my change fit the binarizer on the union of labels (which could introduce an all-zero column). The original fits on `y_true` alone in the multiclass branch, where every class column always has both a positive and a negative, so an empty/degenerate filter can't arise. With the change reverted, the guard isn't needed. (48a05d8b)

🤖 _Automated reply_

validmind/tests/model_validation/sklearn/MinimumF1Score.py
hunner: Fixed — simplified to a single `np.unique` on the concatenated arrays: `len(np.unique(np.concatenate([y_true, y_pred])))`. The `MinimumROCAUCScore` occurrence is gone, since that change was reverted as a verified no-op. (48a05d8b)

🤖 _Automated reply_

hunner:
Thanks for the follow-ups — handling each:

1. **`MinimumROCAUCScore` docstring** — moot: I reverted the `MinimumROCAUCScore` change entirely (it was a verified no-op, see the inline threads), so there's no new exclusion behavior to document.
2. **Shared `_dataset_with_predictions` helper** — left inline per your "fine for now". After the revert only one test file uses it, so there's nothing to deduplicate yet; worth extracting when the next sklearn test adopts the pattern.
3. **Deliberate binary-only guard in `ROCCurve`/`PrecisionRecallCurve`** — added a one-line comment to each. (8e465ff6)

Also FYI: the branch was rebased into two logical commits (the `MinimumF1Score` fix + its tests, and the curve-comment docs) — the `MinimumROCAUCScore` add/revert no longer appears in history. Net diff is strictly smaller than what you approved.

🤖 _Automated reply_

PR #534, opened 2026-07-08
Fix multiclass / non-{0,1}-label crashes across the sklearn classification tests (ZD-704)

johnwalz97:
# Pull Request Description

## What and why?

A customer (ZD-704) running a **multiclass** model hit `ValueError`s in the scikit-learn tests, one at a time. Rather than fix them reactively, I swept **every classification sklearn test** against a multiclass model and fixed all the genuine breakages.

Root cause (shared): these tests score classification metrics / build plots using scikit-learn defaults that only hold for a standard `{0, 1}` binary problem.

### Tests fixed

**Per-region diagnosis tests** — `WeakspotsDiagnosis`, `OverfitDiagnosis`, `RobustnessDiagnosis`. They score precision/recall/F1/AUC on **subsets** of the data (feature bins, noise-perturbed copies):
- precision/recall/F1 default to `average="binary"`, `pos_label=1` → `Target is multiclass but average='binary'` (multiclass) and `pos_label=1 is not a valid label` (binary encoded outside `{0,1}`, e.g. `{0,4}`).
- `roc_auc_score` (default `auc` metric) → `multi_class must be in ('ovo', 'ovr')` on multiclass. A true probability-based multiclass ROC AUC isn't possible here (the library retains only a single probability column), so multiclass AUC falls back to the label-binarize convention already used by `ClassifierPerformance`.

The customer's `WeakspotsDiagnosis` case was a genuinely multiclass target where one feature bin happened to hold only two classes (`0` and `4`) — incidental, not a deliberate encoding (customer confirmed: not binary, not intentionally `0`/`4`).

**`CalibrationCurve`** — wraps sklearn's binary-only `calibration_curve`, which raised a cryptic `pos_label is not specified` on multiclass. Now skips cleanly with `SkipTestError`, matching `ROCCurve`/`PrecisionRecallCurve`.

**`SHAPGlobalImportance`** — `select_shap_values` only handled the older SHAP list layout (one 2D array per class). Newer SHAP returns a single `(samples, features, classes)` array for multiclass, which fell through to the regression branch and was passed whole to `shap.summary_plot` → `IndexError: index N is out of bounds for axis 1`. Now selects along the trailing class axis; multiclass works with `class_of_interest`, and raises a clear, actionable error without it.

## What changed

New shared module `_diagnosis_metrics.py` (`resolve_averaging`/`full_labels`, `bind_averaging`/`apply_averaging`, `multiclass_auc`) used by all three diagnosis tests. `WeakspotsDiagnosis` refactored onto it; `OverfitDiagnosis`/`RobustnessDiagnosis` route `auc` (multiclass) and label metrics through it. Averaging is resolved **once** from the full label set (union of true + predicted labels) so every subset is scored consistently; standard `{0,1}` stays on `pos_label=1` (**unchanged**). Plus the two standalone fixes above.

## Full multiclass sweep (21 classification tests)

| Status | Tests |
|---|---|
| ✅ Fixed here | WeakspotsDiagnosis, OverfitDiagnosis, RobustnessDiagnosis, CalibrationCurve (skips), SHAPGlobalImportance |
| ✅ Already worked | ClassifierPerformance, ConfusionMatrix, MinimumAccuracy, MinimumF1Score, MinimumROCAUCScore, ModelParameters, ModelsPerformanceComparison, FeatureImportance, PermutationFeatureImportance, PopulationStabilityIndex, TrainingTestDegradation, HyperParametersTuning |
| ✅ Binary-by-design, skip gracefully | ROCCurve, PrecisionRecallCurve, ClassifierThresholdOptimization |

`ScoreProbabilityAlignment` needs a scorecard "score" column (not a multiclass issue). Clustering/regression tests aren't applicable to a classifier.

## How to test

```
uv run python -m pytest tests/unit_tests/model_validation/sklearn/ -q
```

New/updated: shared-helper units (incl. `multiclass_auc` and `select_shap_values` edge cases) and end-to-end runs across binary `{0,1}`, binary `{0,4}`, multiclass (default and per-metric), and regression. Full sklearn unit dir: 66 passed.

## What needs special review?

- **Multiclass AUC semantics**: no per-class probabilities are stored, so multiclass `auc` in Overfit/Robustness is a *label-binarize* AUC (matches `ClassifierPerformance`), not a probability ROC AUC.
- Resolving averaging **globally** (from the full dataset) rather than per subset — deliberate, so metrics stay comparable and sparse subsets don't flip to binary.
- `pos_label = max(labels)` for non-`{0,1}` binary targets.
- `SHAPGlobalImportance` now **requires** `class_of_interest` for multiclass (clear error if omitted) — consistent with the pre-existing list-branch contract, but confirm that's the desired UX vs. defaulting to a class.

No breaking changes; `{0,1}` binary and regression results unchanged. Needs a new library release tagged to PyPI before the customer can `pip install` the fix.

## Release notes

Fixed `ValueError`s when running scikit-learn tests on multiclass targets (or binary targets encoded outside `{0, 1}`): `WeakspotsDiagnosis`, `OverfitDiagnosis`, and `RobustnessDiagnosis` now handle multiclass; `CalibrationCurve` skips cleanly for multiclass; and `SHAPGlobalImportance` supports multiclass via `class_of_interest`. Previously these raised errors such as `Target is multiclass but average='binary'`, `pos_label=1 is not a valid label`, `multi_class must be in ('ovo', 'ovr')`, or a SHAP `IndexError`.

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied (`bug`)
- [x] PR linked to Shortcut
- [x] Unit tests added
- [x] Tested locally

### DIFF
--- a/tests/unit_tests/model_validation/sklearn/test_CalibrationCurve.py
+++ b/tests/unit_tests/model_validation/sklearn/test_CalibrationCurve.py
@@ -1,0 +1,43 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+from sklearn.linear_model import LogisticRegression
+
+import validmind as vm
+from validmind.errors import SkipTestError
+from validmind.tests.model_validation.sklearn.CalibrationCurve import CalibrationCurve
+
+
+def _dataset_and_model(labels, seed):
+    n = 200
+    rng = np.random.RandomState(seed)
+    y = rng.choice(labels, size=n)
+    df = pd.DataFrame(
+        {"f1": np.linspace(0.0, 10.0, n), "f2": rng.randn(n), "target": y}
+    )
+    ds = vm.init_dataset(
+        input_id=f"cc_{seed}", dataset=df, target_column="target", __log=False
+    )
+    est = LogisticRegression(max_iter=1000).fit(df[["f1", "f2"]].to_numpy(), y)
+    model = vm.init_model(input_id=f"cc_{seed}_m", model=est, __log=False)
+    ds.assign_predictions(model=model)
+    return ds, model
+
+
+class TestCalibrationCurve(unittest.TestCase):
+    def test_multiclass_skips_cleanly(self):
+        # sklearn's calibration_curve is binary-only and raised a cryptic
+        # "pos_label is not specified" error on multiclass; it must skip instead.
+        ds, model = _dataset_and_model([0, 1, 2, 3, 4], seed=1)
+        with self.assertRaises(SkipTestError):
+            CalibrationCurve(model=model, dataset=ds)
+
+    def test_binary_runs(self):
+        ds, model = _dataset_and_model([0, 1], seed=2)
+        result = CalibrationCurve(model=model, dataset=ds)
+        self.assertIsNotNone(result)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/unit_tests/model_validation/sklearn/test_OverfitDiagnosis.py
+++ b/tests/unit_tests/model_validation/sklearn/test_OverfitDiagnosis.py
@@ -1,0 +1,76 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+from sklearn.linear_model import LinearRegression, LogisticRegression
+
+import validmind as vm
+from validmind.tests.model_validation.sklearn.OverfitDiagnosis import OverfitDiagnosis
+
+
+def _pair(labels, seed=0, n=200, classification=True):
+    """Build a train/test VMDataset pair sharing one fitted VMModel."""
+
+    def build(input_id, s, model=None):
+        r = np.random.RandomState(s)
+        f1 = np.linspace(0.0, 10.0, n)
+        f2 = r.randn(n)
+        if classification:
+            y = r.choice(labels, size=n)
+            # ensure the lowest feature bin holds only a subset of the classes
+            y[:20] = r.choice(labels[:2], size=20)
+        else:
+            y = 3.0 * f1 + 2.0 * f2 + r.randn(n)
+        df = pd.DataFrame({"f1": f1, "f2": f2, "target": y})
+        ds = vm.init_dataset(
+            input_id=input_id, dataset=df, target_column="target", __log=False
+        )
+        if model is None:
+            est = (
+                LogisticRegression(max_iter=2000)
+                if classification
+                else LinearRegression()
+            )
+            est.fit(df[["f1", "f2"]].to_numpy(), y)
+            model = vm.init_model(input_id=f"{input_id}_m", model=est, __log=False)
+        ds.assign_predictions(model=model)
+        return ds, model
+
+    train, model = build(f"of_{seed}_train", seed)
+    test, _ = build(f"of_{seed}_test", seed + 1, model=model)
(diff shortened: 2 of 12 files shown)


hunner (approved):
Approving — this subsumes #529's follow-up work in #533 (which I'm closing in favor of this PR) and I verified the claims empirically rather than by read-through: a 15-scenario matrix (Weakspots / Overfit / Robustness × default `auc` / explicit `f1` × labels `{0,2,4}` / `{0,4}` / `{0,1}`) run identically against `main`, #533, and this branch. `main` crashes on 9 of 15; this branch passes all 15, including the multiclass default-`auc` path that #533 left out of scope (and which would very likely have been this customer's next ticket). `multiclass_auc` does match `ClassifierPerformance`'s LabelBinarizer convention as claimed, with sensible hardening on top (global-label fit for column alignment, scorable-class filtering; unseen perturbed labels binarize to all-zero rows rather than crashing). Local suite: 58 passed, 2 pre-existing skips (SHAP tests covered by the max-deps CI matrix, which is green).

Two findings worth fixing before merge, both demonstrated:

**1. `CalibrationCurve` still crashes on binary targets encoded outside `{0, 1}` — the exact error this PR calls out as cryptic.** The new guard only skips when `len(np.unique(dataset.y)) > 2`, so a two-class `{0, 4}` target sails past it and hits sklearn's `ValueError: y_true takes value in {0, 4} and pos_label is not specified...`. Reproduced on this branch with a `{0, 4}` logistic model. Since the PR's own convention (in `resolve_averaging`) is "largest label is the positive class", passing `pos_label` through to `calibration_curve` for the two-class case closes it consistently:

```python
prob_true, prob_pred = calibration_curve(
    dataset.y, dataset.y_prob(model), n_bins=n_bins,
    pos_label=np.unique(dataset.y).max(),
)
```

**2. `apply_averaging` in `WeakspotsDiagnosis` rebinds user-supplied custom metrics, silently overriding their explicit choices.** `bind_averaging` rebinds any callable whose signature exposes `average` — and `inspect.signature` of `functools.partial(f1_score, average="weighted")` still exposes `average`, so the most natural way for a user to customize averaging gets clobbered. Demonstrated on this branch:

```python
p = functools.partial(f1_score, average="weighted")
bind_averaging(p, "macro", None)(y_true, y_pred)
# returns the macro score (0.8413), not the user's weighted score (0.8492)
```

Since `WeakspotsDiagnosis` applies it to the whole prepared registry, this hits every custom metric, not just defaults. Suggest only applying averaging when the caller didn't pass `metrics` (i.e. rebind `DEFAULT_METRICS` only) — custom callables should own their kwargs.

Two non-blocking notes: the PR body's "every subset is scored consistently" slightly oversells — the *strategy* is resolved globally but macro without `labels=` still averages over each slice's present classes only (a `{0,2}` slice of a `{0,2,4}` problem: 0.80 here vs 0.53 if `labels=` were pinned). I actually think your behavior is the better choice for this test — a homogeneous, perfectly-predicted bin shouldn't be flagged as a weakspot — so just worth a precise sentence rather than a change. And the Shortcut checklist item: this is [sc-17091](https://app.shortcut.com/validmind/story/17091).

🤖 _Automated reply_

PR #535, opened 2026-07-09
[SC-17128] ZD-717: Add multiclass (one-vs-rest) support to VM tests

AnilSorathiya:
# Pull Request Description

## What and why?
Adds **multiclass (one-vs-rest) support** to four sklearn model-validation tests that were
previously binary-only, so customers validating multiclass classifiers get usable results
instead of a skipped/binary-only test (requested via ZD-717).

Per-class probabilities are read from the underlying estimator's `predict_proba` — the
`VMModel` wrapper only exposes the positive-class column, which is binary-only — and the true
labels are one-hot encoded with `label_binarize` in sklearn's sorted-class order to stay
aligned with the `predict_proba` columns.

- **ROCCurve** — for >2 classes, plots one OvR ROC curve per class plus a micro-average, each
  with its own AUC; `RawData` returns per-class `fpr`/`tpr`/`auc` keyed by class label plus a
  `micro` entry. Docstring updated for multiclass.
- **PrecisionRecallCurve** — per-class OvR precision-recall curves plus a micro-average.
- **ConfusionMatrix** — for >2 classes uses the model's argmax class predictions; the
  `threshold` parameter is binary-only and is ignored for multiclass.
- **PopulationStabilityIndex** — computes PSI one-vs-rest, one table/plot per class.

Binary behavior is unchanged in all four. Where a full per-class probability matrix isn't
available (metadata-only models, precomputed single-column probabilities, or a shape that
doesn't match the class count), the multiclass path raises a clear `SkipTestError` rather than
crashing.

### Automated tests
Run the added/updated unit tests (this repo runs unittest via `uv`):

```bash
make test ONLY="tests.unit_tests.model_validation.sklearn.test_ROCCurve tests.unit_tests.model_validation.sklearn.test_PrecisionRecallCurve"
```

New multiclass coverage: `TestROCCurveMulticlass.test_multiclass_one_vs_rest_traces` and
`TestPrecisionRecallCurveMulticlass.test_multiclass_one_vs_rest_traces` (both train a 3-class
XGBoost model and assert one curve per class + a micro-average trace + baseline, and per-class
AUCs above 0.5). Binary tests in the same files should continue to pass.

> Note: the multiclass tests require the `xgboost` optional extra and are skipped without it.

### Manual testing
Run `notebooks/code_samples/multiclass_classification_tests.ipynb` end-to-end against a
3-class model and confirm:
- ROC and PR plots show one curve per class plus a micro-average curve.
- The confusion matrix renders an N×N grid from argmax predictions.
- PSI renders one table/plot per class.
- Re-running with a binary model produces the original single-curve output.

## What needs special review?
- **Label/column alignment**: `label_binarize(y, classes=...)` must match `predict_proba`
  column order (sklearn's sorted class labels). Misalignment would silently mislabel per-class
  curves.
- **Skip conditions**: the `SkipTestError` fallbacks should skip only genuinely unsupported
  models (no per-class `predict_proba`, wrong-shaped probability matrix) and not falsely skip
  supported ones.
- **ConfusionMatrix**: `threshold` is now ignored for multiclass — confirm this is the desired
  behavior and is clearly documented.

None. Library-only change, no migrations, no new env vars, no companion frontend/backend PR.
Binary output is unchanged, so no breaking changes.

Shortcut: [SC-17128](https://app.shortcut.com/validmind/story/17128) ·

## Release notes

ROC, Precision-Recall, Confusion Matrix, and Population Stability Index tests now support
multiclass classification models, in addition to binary. For models with more than two
classes, ROC and Precision-Recall curves are drawn one-vs-rest (one curve per class plus a
micro-average), the confusion matrix uses the model's predicted class, and PSI is reported per
class.

## Checklist
- [x] What and why
- [ ] Screenshots or videos (Frontend) — N/A (no UI change)
- [x] How to test
- [x] Labels applied — `enhancement`
- [x] PR linked to Shortcut
- [x] Tested locally — `test_ROCCurve` + `test_PrecisionRecallCurve` pass (5 tests, incl. multiclass OvR)


cachafla (approved):
Tested the notebook and is working great 👏 

If we're going to distribute this notebook I suggest:
- Ensuring it has the same structure as other public notebooks (intro text, code snippet, footer, etc.
- Put it inside a directory, right now it's at the root of `code_samples/`

Otherwise, we can keep it inside `code_sharing/`.

AnilSorathiya:
> Tested the notebook and is working great 👏
> 
> If we're going to distribute this notebook I suggest:
> 
> * Ensuring it has the same structure as other public notebooks (intro text, code snippet, footer, etc.
> * Put it inside a directory, right now it's at the root of `code_samples/`
> 
> Otherwise, we can keep it inside `code_sharing/`.

It is not something for demonstration. We can keep in inside `code_sharing`.

johnwalz97 (approved):
lgtm

PR #546, opened 2026-07-23
Hunner/release 2.13.8

hunner:
# Pull Request Description

## What and why?
Releasing v2.13.8

Includes
- #538 
- #540

## Release notes
None; just a release bump pr

## Checklist
- [x] What and why
- [ ] How to test
- [x] Labels applied
- [ ] PR linked to Shortcut
- [ ] Tested locally


juanmleng:
Nice catch on the Makefile fix. Switching `sed -i ''` (which only works on BSD/macOS sed) to the `sed '...' > .tmp && mv` form means `make version` will now work in Linux CI rather than only on a Mac, and the `&&` keeps the original file intact if `sed` ever fails.

One tiny thing, take it or leave it: if `sed` does fail, the `DESCRIPTION.tmp` gets left behind. Since the target does an explicit-file `git add`, it won't actually get committed, so this is really just tidiness — a `|| rm -f r/validmind/DESCRIPTION.tmp` would clean it up if you want it airtight.

LGTM 👍

juanmleng (approved):

PR #548, opened 2026-07-24
fix: refresh OIDC access token on request path

hunner:
# Pull Request Description

## What and why?

When authenticating with OIDC (device flow), the ValidMind library captured the access token exactly once, inside `vm.init()`, and reused it verbatim for the entire life of the process. `_get_api_headers()` read the cached token unconditionally — there was no expiry check and no refresh anywhere on the request path. All of the refresh machinery already existed (`try_refresh_cached_tokens`, `is_expired` with a 120s skew), but it was only reachable from `init()`.

**Before:** a long-running session (notebook, batch job) that outlived the provider's access-token lifetime — Okta's default is 1 hour — started failing on *every* subsequent API call once the token expired. The only recovery was manually re-running `vm.init()`. Worse, the failure surfaced as a generic `APIRequestError` carrying the backend's description, with no hint that the token had expired, so users couldn't self-diagnose (the reporter found the cause by reading our source).

**After:** the token is refreshed on the request path. `_get`/`_post`/`_ping` proactively refresh via the cached refresh token when the current token is within the expiry skew, so a valid bearer is attached before the request goes out. `_get`/`_ping` additionally force one refresh-and-retry if a `401` still comes back (e.g. a token revoked ahead of its stated expiry). A genuinely rejected OIDC token now raises a clear `ValidMindAuthError` telling the user to re-run `vm.init()`.

This depends on a refresh token being available, which is why it pairs with #538 (adds `offline_access` to the default scope, shipped in v2.13.8). #538 alone let a *manual* `vm.init()` re-run refresh silently; this change makes refresh happen transparently mid-session, which is the behavior the ticket actually asked for.

## How to test

1. Authenticate against an OIDC provider that issues short-lived access tokens plus a refresh token (default scope now includes `offline_access`):
   ```python
   import validmind as vm
   vm.init(issuer="https://<your-okta>/oauth2/default", client_id="<cid>", model="<cuid>")
   ```
2. Wait past the access-token lifetime (or configure a short one), then make any API call (e.g. `vm.log_metadata(...)` or anything that hits `_get`/`_post`). Previously this failed; it should now transparently refresh and succeed, with `~/.validmind/credentials.json` showing an updated `expires_at`.
3. Revoke the token server-side mid-session and make a call: `_get`/`_ping` should force-refresh and retry once; if refresh is impossible, you should get a clear `ValidMindAuthError` mentioning `vm.init()` rather than a generic error.
4. `.venv/bin/python -m unittest tests.test_api_client` — 37 tests pass, including new ones covering proactive refresh, forced refresh, the no-refresh-token path, api-key no-op, the clear 401 error, the stampede-guard double-check, drop-cache-on-refresh-failure, and end-to-end `_get`/`_post`/`_ping` refresh-and-retry wiring.

## What needs special review?

The main review surface is **shared mutable auth state under concurrency**. `_access_token` is a module global and the aiohttp session bakes the bearer header at construction, so refresh has to mutate the global *and* invalidate the pooled session (`_set_oidc_access_token`). Points worth a close read:

- **Refresh serialization.** Refresh runs under `_oidc_refresh_lock` (a `threading.Lock`) with a double-check inside the lock, so concurrent callers that all see an expired token don't stampede the token endpoint — the first refreshes, the rest observe the freshly-cached token. The lock is a plain `threading.Lock` rather than an `asyncio.Lock` because `_ensure_fresh_oidc_token` performs no `await` while held (the refresh HTTP call is synchronous `requests`), so it is never held across an await point; it exists to serialize refreshes across threads (e.g. multiple `run_async` callers). The trade-off is that the synchronous refresh briefly blocks the event loop — acceptable given it happens roughly once per token lifetime.
- **Asymmetric 401 handling.** `_get` and `_ping` (bodyless GETs) force-refresh and retry once on a 401. `_post` deliberately does **not** retry, because the request body / upload stream may already be consumed — replaying it isn't safe. For `_post`, proactive refresh covers the expiry case, and a residual 401 raises the clear auth error. If reviewers want transparent retry on writes too, that needs body/stream re-materialization (e.g. `seek(0)` on upload `BytesIO`) and should be a follow-up.
- **Hot-path cost.** The common (token-valid) path is a cheap in-memory expiry check against `_oidc_expires_at` — no file I/O, no lock. File reads and the lock only happen when the token is actually stale.

Complements #538 (`offline_access` default scope, v2.13.8) — refresh requires a cached refresh token, which that scope enables. No breaking changes; API-key auth is entirely unaffected (all new logic is gated on `_auth_mode == "oidc"`). No deployment considerations.

## Release notes

`bug`

The ValidMind library now automatically refreshes an expired OIDC access token during a session instead of failing every API call until `vm.init()` is re-run, and reports a clear, actionable error when a token is rejected.

## Checklist
- [x] What and why
- [x] How to test
- [ ] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally

Closes ZD #682 / [sc-16892](https://app.shortcut.com/validmind/story/16892/zd-682-vm-python-library-oidc-access-token-not-automatically-refreshed)

### DIFF
--- a/tests/test_api_client.py
+++ b/tests/test_api_client.py
@@ -281,7 +281,9 @@ class TestAPIClient(unittest.TestCase):
             200,
             json={
                 "content_id": "dataset_summary_text",
-                "text": md_to_html("## Generated Summary\nGenerated content.", mathml=True),
+                "text": md_to_html(
+                    "## Generated Summary\nGenerated content.", mathml=True
+                ),
             },
         )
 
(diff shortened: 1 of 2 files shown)


jamadriz (changes requested):
looks great, minor gaps that would suggest covering in this PR (last one ok to defer to followup PR):

1. Sync requests.post paths are still uncovered

The PR hooks `_get`, `_post`, and `_ping`, but these sync callers still go straight to requests with no refresh:

api_client.py
Lines 657-667
```python
def generate_qualitative_text(text_generation_data: Dict[str, Any]) -> Dict[str, Any]:
    """Generate qualitative text using the ValidMind AI API."""
    r = requests.post(
        url=_get_url("ai/generate/qualitative_text_generation"),
        headers=_get_api_headers(),
        json=text_generation_data,
    )
```
Same pattern for `generate_test_result_description`. A notebook that only hits AI generation endpoints after token expiry could still hit the old failure mode. Suggest either `adding _ensure_fresh_oidc_token()` there or extracting a small shared “make authenticated request” wrapper.

2. Refresh-failure behavior differs from `init()` path

In `_obtain_oidc_tokens`, refresh failure deletes the cache entry and falls back to device flow. In `_ensure_fresh_oidc_token`, refresh failure returns False and leaves cache intact:
```python
except ValidMindAuthError:
    return False
```
That’s defensible for transient errors, but on `invalid_grant` (revoked refresh token) every request will retry refresh until a 401 surfaces. Aligning with `init`’s `delete_cached_entry` on hard failure would be cleaner — or at least document the intentional difference.

3. Tests cover the helper, not the wiring 

The six new tests exercise `_ensure_fresh_oidc_token` well, but none assert that `_get` / `_post` / `_ping` actually call it or that 401 retry works end-to-end. Low risk given the thin wiring, but one integration-style test per path would strengthen confidence.

4. Optional DRY (not required)

The refresh block in `_ensure_fresh_oidc_token` mirrors the refresh branch in `_obtain_oidc_tokens`. A shared `_refresh_oidc_from_cache(...)` helper could reduce drift later, but that’s a follow-up — not redundancy today.

hunner:
All good points. Updated. One thing is that the request-path refresh can't fall back to the device flow if the refresh fails like `init()` can, so that just bails.

jamadriz (approved):
looks good 🚀

jamadriz:
###  Remaining nits (non-blocking)
* PR description is stale — still says "32 tests"; the branch now has 37. Worth a one-line update before merge.
* AI-gen endpoints lack 401 retry — same intentional constraint as _post (no safe body replay). Proactive refresh covers normal expiry; residual 401 gets the clear error. Fine as-is.
* In-memory token not cleared on refresh failure — after delete_cached_entry, _access_token / _oidc_expires_at still hold stale values until vm.init(). Harmless since the next request 401s with a clear message, but clearing in-memory state on hard failure would be slightly cleaner (optional follow-up).

